PR #5, opened 2021-11-14
test: add YaruExtraOptionRow test

shan-shaji:


Feichtmeier:
@shan-shaji thank you very much! And sorry again for not answering for so long

PR #16, opened 2021-12-05
docs: add yaruCheckBoxRow docs

shan-shaji:


Feichtmeier:
Thank you! Nice idea
We should add more docs to other devices

shan-shaji:
Yes, that would be nice

PR #20, opened 2021-12-07
Add Github Action Flutter CI

Jupi007:
Because each good repo should have some CI actions ;)

@Feichtmeier I added the `flutter format` check but I'm not sure. What do you think of?


Jupi007:
Also maybe use `flutter analyze --fatal-infos`?

Feichtmeier (approved):
Thanks!
I think fatal Infos is not needed

PR #22, opened 2021-12-09
Add YaruCheckbox Material based widget

Jupi007:
Add a new `YaruCheckbox` widget, fully compatible with the material one.

https://user-images.githubusercontent.com/36476595/145389440-d00dde0a-ea1c-400c-905b-e68e467a8267.mp4

I copied the code of the material `Checkbox` and I tweaked the icon drawing function, and the size. You can see all changes I did with `// Yaru change` comments (just like the Yaru repository, I'll add later an action which will check upstream change).

The benefit of tweaking the code upstream is:

- it is fully compatible with the material `Checkbox`, and so we can replace each previous with a `YaruCheckbox` without any problem ;
- we keep the same look and feel as the material `Checkbox`: check animation, splash, ...

### DIFF
--- a/example/lib/yaru_master_details_page.dart
+++ b/example/lib/yaru_master_details_page.dart
@@ -13,6 +13,7 @@ class YaruHome extends StatefulWidget {
 }
 
 class _YaruHomeState extends State<YaruHome> {
+  bool? _checkValue = false;
   bool _extraOptionValue = false;
   bool _isImageSelected = false;
   final TextEditingController _textEditingController = TextEditingController();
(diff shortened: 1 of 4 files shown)


Feichtmeier (approved):
Wow, that looks like a bunch of work, ty!

Feichtmeier:
a wait

Feichtmeier (changes requested):
Wouldn't it be possible to extends from the material checkbox and only overwriting the method that draws the check?

Jupi007:
> Wouldn't it be possible to extends from the material checkbox and only overwriting the method that draws the check?

I tried, but I can't. 😔 It could be indeed really much simpler.
I can extends the `Checkbox` class, but as it's a stateful class, it needs to overwrite the `_CheckboxState`, which is private and not extendable.

PR #23, opened 2021-12-09
Add YaruSwitch widget

Jupi007:
Add a new `YaruCheckbox` widget, fully compatible with the material one.

Capture d’écran du 2021-12-09 21-49-40
Capture d’écran du 2021-12-09 21-49-49

I copied the code of the material `Switch` and I tweaked some size const. You can see all changes I did with `// Yaru change` comments (just like the Yaru repository, I'll add later an action which will check upstream change).

I also bypassed all `CupertinoSwitch` stuff with comments.

**Note:** imo, yaru switch colors looks bad. If you agree @Feichtmeier I'd like to make a PR at `yaru.dart` to make them using the same colors as Gtk4.

### DIFF
--- a/example/lib/yaru_master_details_page.dart
+++ b/example/lib/yaru_master_details_page.dart
@@ -14,6 +14,7 @@ class YaruHome extends StatefulWidget {
 
 class _YaruHomeState extends State<YaruHome> {
   bool? _checkValue = false;
+  bool _switchValue = false;
   bool _extraOptionValue = false;
   bool _isImageSelected = false;
   final TextEditingController _textEditingController = TextEditingController();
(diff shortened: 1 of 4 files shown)


Feichtmeier:
Nice work!
It should be possible to change this knob here without yaru.dart
I think yaru.dart should stay 100% material

Jupi007:
I can try to hardcode colors, but it will break the compatibility with material theme api.
On the other hand if I change color in `yaru.dart` we keep material api compatibility for `YaruSwitch` but material `Switch` colors will no longer be appropriate.

Let's keep material switches for Ubuntu apps. They look not bad at all. And they don't need more branding tweaks like checkboxes (which was really miss our icon).

What do you think?

Feichtmeier:
So it's possible...
grafik

But after thinking about this for a while I think it is weird to have two widgets almost like material with some small changes but with a lot of code.
I think in dart there is the possibility to create an "extension" on a class
https://dart.dev/guides/language/extension-methods
Maybe this could be used for both the new switch and the new check to minimize the additional code?

Jupi007:
> But after thinking about this for a while I think it is weird to have two widgets almost like material with some small changes but with a lot of code.
> I think in dart there is the possibility to create an "extension" on a class
> https://dart.dev/guides/language/extension-methods
> Maybe this could be used for both the new switch and the new check to minimize the additional code?

It's indeed really annoying to have these huge files for such small modifications, but there's no other solution, and that really bothers me a lot. The problem is that we can't extends (or make an extension) on a private class. If you search on internet how to extends a material widget, everybody said this is impossible. 😔

About `yaru.dart` styling, I wanted to do very soft modifications:

Capture d’écran du 2021-12-10 11-01-56
Capture d’écran du 2021-12-10 11-02-11
Capture d’écran du 2021-12-10 11-04-43
Capture d’écran du 2021-12-10 11-05-00

Like this we don't break anything in the material theming api. And it looks good with both Yaru material and full material widgets.

Feichtmeier:
I don't think it looks good like this. The background and knob combination somehow doesn't fit to the rest of the UI. Maybe it's because I've looked onto the material Yaru for so long but I think we shouldn't do it. 
Soon there is some second switch style in flutter, it's already in normal Android 12

Maybe we should just wait for it and also revert the checkbox? The idea looked good on paper but I think we shouldn't start mixing those basic widgets :(

Issue #26, opened 2021-12-20
Create a tabbed view widget


Feichtmeier:
The default tab styling is a bit unsatisfying. 

Created a tabbed view which makes sense for different layers on a page, like for example in the settings app in the connections page:
![grafik](https://user-images.githubusercontent.com/15329494/146746478-a71d5c35-0514-42a9-bcef-6e3225f61f96.png)

and the keyboard page:
![grafik](https://user-images.githubusercontent.com/15329494/146746548-48f5a0ca-1038-4967-bde1-0a9a95b43074.png)

- tab count
- list of titles
- list of icons
- list of view widgets
- and a better idea than hardcoded height "1000" :roll_eyes: :laughing: 

```dart
class KeyboardPage extends StatefulWidget {
  const KeyboardPage({Key? key}) : super(key: key);

  static Widget create(BuildContext context) {
    return const KeyboardPage();
  }

  @override
  State<KeyboardPage> createState() => _KeyboardPageState();
}

class _KeyboardPageState extends State<KeyboardPage>
    with TickerProviderStateMixin {
  late TabController tabController;

  @override
  void initState() {
    tabController = TabController(length: 2, vsync: this);
    super.initState();
  }

  @override
  Widget build(BuildContext context) {
    return Column(
      children: [
        Container(
          width: 516,
          height: 60,
          decoration: BoxDecoration(borderRadius: BorderRadius.circular(4)),
          child: TabBar(
              controller: tabController,
              indicator: BoxDecoration(
                  borderRadius: BorderRadius.circular(4),
                  color:
                      Theme.of(context).colorScheme.onSurface.withOpacity(0.1)),
              tabs: const [
                Tab(
                    icon: Icon(YaruIcons.input_keyboard),
                    child: Text("Keyboard Settings")),
                Tab(
                    icon: Icon(YaruIcons.font),
                    child: Text("Keyboard Shortcuts")),
              ]),
        ),
        Padding(
          padding: const EdgeInsets.only(top: 30),
          child: SizedBox(
            height: 1000,
            child: TabBarView(
              controller: tabController,
              children: const [KeyboardSettingsPage(), KeyboardShortcutsPage()],
            ),
          ),
        ),
      ],
    );
  }
}
```

PR #31, opened 2021-12-21
YaruTabbedPage

Feichtmeier:
\!\[ya\]\(https://user\-images.githubusercontent.com/15329494/147101579\-fd624d8a\-a9ab\-4763\-b479\-fff4272fc14e.gif\)

TODO:

- [X] improve tab hover/active
- [X]  add docs
- [X] do not hardcode the height

Closes \#26

### DIFF
--- a/example/lib/main.dart
+++ b/example/lib/main.dart
@@ -1,6 +1,6 @@
 import 'package:flutter/material.dart';
 import 'package:yaru/yaru.dart' as yaru_theme;
-import 'package:yaru_widgets_example/yaru_master_details_page.dart';
+import 'package:yaru_widgets_example/yaru_home.dart';
 
 void main() {
   runApp(Home());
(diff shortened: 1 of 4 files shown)


Feichtmeier:
@Jupi007 with the material tabbar style it is impossible to have a rounded inkwell on hover/pressed
I tried to just remove them, do you think this is better or worse than having a not-rounded hover/pressed?

Jupi007:
Imo, it is better to drop the ripple effect to keep border-radius.

Btw, this looks like an issue of flutter itself: https://github.com/flutter/flutter/issues/50341

Also, we still see ripple when doing a long click on far right or left tab.
Instead of using a `Container`, you should wrap the `TabBar` inside a `Theme` like this:

```dart
Theme(
  data: ThemeData().copyWith(
    splashColor: Colors.transparent,
    hoverColor: Colors.transparent,
    highlightColor: Colors.transparent,
  ),
  child: TabBar(
    ...
))
```

Feichtmeier:
Ah nice idea =)
I cheated out the hover and pressed color to transparent with indicatorColor: MaterialStateProperty.resolveWith(states....
but this looks better, will do this!

PR #36, opened 2022-01-01
Decrease YaruSimpleDialog close button splashRadius

### LINKED ISSUE #32: YaruSimpleDialog: close button should be inside the dialog frame
grafik

I think the close button should be inside a "regular" 4px border radius button and stay inside the dialog
or just be a smaller circle with enough padding so the position of the X stays the same but the circle does not leave the frame

WDYT @Jupi007 ?

Jupi007:
Decrease YaruSimpleDialog close button `splashRadius` to 16.

Capture d’écran du 2022-01-01 22-15-05

Closes #32


Feichtmeier (approved):
Tyvm!

PR #42, opened 2022-01-07
Disable instead hide

mivoligo:
As discussed here: https://github.com/Feichtmeier/settings/issues/1982 I made some changes to allow widgets to look disabled instead of displaying empty SizedBoxes when there's no value.


Feichtmeier (approved):
Thank you very much!

Issue #41, opened 2022-01-06
ImageTile: expose filter quality


Feichtmeier:
The filter quality of the image child needs to be accessible via the ImageTile constructor.

Feichtmeier:
Fixed by #48

PR #48, opened 2022-01-13
fix: exposed filterQuality through imageTile(#41)

shan-shaji:


Feichtmeier:
Hi @shan-shaji !

Thank you very much!

PR #49, opened 2022-01-14
Remove secondary slider

mivoligo:
As far as I know the secondary slider widget was only used by me in some dialogs in a11y page in the Settings app. It was like a regular Yaru slider but with height specified. I believe we don't need it anymore.


Feichtmeier:
Alright :) I was always confused what the difference was but never dared to touch it xD

PR #63, opened 2022-02-01
Dialog update

Feichtmeier:
- extract the title to an individual widget
- add a YaruAlertDialog


Jupi007:
Very nice idea to extract the dialog title :D

Feichtmeier:
It was @jpnurmi idea :brain: :smile:

Feichtmeier:
@Jupi007  could you check if you see any behavioral changes to the previous version which you introduced?

grafik
grafik

I also thought that the alert dialog should have a left aligned title - wdyt?

Jupi007 (approved):
LGFM :+1: 
I didn't see any regression on my side.
And the left aligned `Alert` title is also very nice :D

PR #69, opened 2022-02-02
YaruImageTile: rename and make it more flexible

### LINKED ISSUE #68: YaruImageTile: replace path with an image widget
A `String path` property is problematic because it doesn't let apps specify the desired type of image widget. For example, the Ubuntu Desktop Installer needs to pass a `FlavorImage` widget for looking up flavor theme thumbnails in canonical/ubuntu-desktop-installer#656. Please consider changing the property to a usual `Widget child` as it is in 95% of cases in Flutter. This would give apps full control over the image.

Feichtmeier:
YaruImageTile is now YaruSelectableContainer

Everything can be put inside, not only images.
Thanks to @jpnurmi for the suggestion

yayayaya

(the little yellow disk comes from the program to record gifs :))

Closes #68


lib/src/yaru_selectable_container.dart
```diff
@@ -0,0 +1,58 @@
...
+  /// else if the value is `false` the border color will be [Colors.transparent].
+  final bool currentlySelected;
+
+  /// Callback triggered when the [YaruSelectableContainer] is clicked.
+  final VoidCallback? onTap;
+
+  /// Optional custom radius for the corners which defaults to a circular 6.0 [BorderRadius]
+  final BorderRadius? borderRadius;
+
+  /// Optional custom padding for the child which defaults to 6.0 on all sides.
```
mivoligo, line 30: Maybe it'd be clearer if the default value was specified in the constructor?

Feichtmeier: ah, well if we would stick with double yes. But I thought it would be better to leave the full freedom of using any combination of different roundings at the 4 edges as one wishes. With the BorderRadius as one property I can not default to BorderRadius.circular() because: "The default value of an optional parameter must be constant."

Feichtmeier: nvm, figured it out myself ^^

mivoligo: Yes, `BorderRadius.cicular()` has no `const` constructor. You can use: `BorderRadius.all(Radius.circular(6))` but I see you managed to do it in a different way :)

PR #101, opened 2022-02-23
Bugfix/example carousel page selectable container page image asset path

defuncart:
When running the example app locally on macos, I noticed that `CarouselPage` and `SelectableContainerPage` threw a `FileSystemException`:

```
Cannot open file, path = 'assets[/ubuntuhero.jpg]()' (OS Error: No such file or directory, errno = 2)
```

Bildschirmfoto 2022-02-23 um 17 28 05

By using `AssetImage`, the example is now correctly working locally for me on web and macOS (through IDE or terminal), and when running a web release build on a local server (i.e. `flutter build web --release`, `npx http-server`).

I have not tried the example on Linux, however I would expect that the existing code also would not work.

BTW it may be an idea to add macOS as an example target. I am happy to open this in another PR if you'd like :)

### DIFF
--- a/example/lib/pages/carousel_page.dart
+++ b/example/lib/pages/carousel_page.dart
@@ -1,6 +1,3 @@
-import 'dart:io';
-
-import 'package:flutter/foundation.dart';
 import 'package:flutter/material.dart';
 import 'package:yaru_icons/yaru_icons.dart';
 import 'package:yaru_widgets/yaru_widgets.dart';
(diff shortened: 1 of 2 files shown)


Feichtmeier:
thank you!
can't find any problem on linux / web locally so let's merge - or are there more commits comming? just asking because you are still committing :)

defuncart:
@Feichtmeier No more commits coming :)

PR #102, opened 2022-02-24
Feature/yaru master detail page remove unused variables

defuncart:
While using `YaruMasterDetailPage`, I was a little confused what `searchHint` etc. were for. Then after looking at the code, I realized they are probably leftovers after being extracted to `YaruSearchAppBar` in the past.


Feichtmeier (approved):
Thanks!
Those members are artifacts from when the search was hardcoded :face_exhaling: :smile:

Issue #104, opened 2022-03-01
Change the content padding of the `YaruAlertDialog`


GroovinChip:
Currently, the `contentPadding` property of the `AlertDialog` that is build by the `YaruAlertDialog` widget is set to `EdgeInsets.zero`. This results in dialogs with content that run right up to the left and right edges, as seen in the screenshot below.

It would be great if you guys could update this property with some default padding. Alternatively, you could expose this property for developers to set themselves.

![Screen Shot 2022-03-01 at 10 47 03 AM](https://user-images.githubusercontent.com/4250470/156200985-552f9af2-abc8-4226-97db-38de6ab47d8e.png)

Jupi007:
@GroovinChip did #107 fixed your problem?

GroovinChip:
I haven’t used the package since I filed this issue as my requirements changed shortly afterwards. I’m going to assume it does fix the problem.

Jupi007:
Okay, thanks :)

PR #107, opened 2022-03-18
Allow custom content padding

mivoligo:
Possible fix for #104


Feichtmeier (approved):
huhhhh why didn't I see this :O Ty!

Issue #106, opened 2022-03-18
Make the Ubuntu logo a widget!


ItzSwirlz:
Just like how the Flutter Logo is. As soon as we get the new updated logo I'd like to do this :)

PR #108, opened 2022-03-21
Add ubuntu logo as a widget (#106)

ItzSwirlz:
It seems the example had issues trying to find the ubuntu_logo.png file. I mean, if you wanted we could manually create the logo from ground up like Flutter does.. (https://github.com/flutter/flutter/blob/master/packages/flutter/lib/src/painting/flutter_logo.dart) but I figured it was no big concern as it was in the example.


mivoligo:
I believe the new Ubuntu logo should be inside a rectangle.

ItzSwirlz:
> I believe the new Ubuntu logo should be inside a rectangle.

I took the circle and cropped it like the normal Ubuntu logo is- but I'm not sure if the new logo with the rectangle (That looks weird and unpositioned) and used it as I felt that image was a general branding. The circle of friends is the big idea here.

However, we should have widgets for the *full* Ubuntu logo information with branding text and whatnot. But first, let's have a review.

ItzSwirlz:
I'm so freaking stupid. I forgot to add the assets to the pubspec.yaml file. I'll do that at home and if it works I'll add an example page and keep going with color variants of the logo

PR #117, opened 2022-05-12
YaruCarousel: allow more actions

Jupi007:
Allow to navigate into a `YaruCarousel` by clicking on a page or a circle.

https://user-images.githubusercontent.com/36476595/168139706-56d9b3fe-b27c-44ef-b1f2-4ec4a7a16481.mp4

### DIFF
--- a/lib/src/yaru_carousel.dart
+++ b/lib/src/yaru_carousel.dart
@@ -1,5 +1,8 @@
 import 'package:flutter/material.dart';
 
+const _kAnimationDuration = Duration(milliseconds: 500);
+const _kAnimationCurve = Curves.easeInOutCubic;
+
 class YaruCarousel extends StatefulWidget {
   const YaruCarousel({
     Key? key,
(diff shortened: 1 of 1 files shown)


Feichtmeier:
Hmmm I don't get this one.
You want to have an ontap for one of the elements in the carousel?
If yes it is not needed since all kinds of widgets are allowed as children. 
But maybe I misunderstood

Jupi007:
Well actually, you can only change the current page by doing a click-slide, which is very not usual on the desktop.
That I propose is to be able to click on one of those bottom circle, or click on the next/previous card which is a bit visible to navigate to the corresponding page.
Ex: if I click on the fourth circle, it navigate to the fourth page.

Feichtmeier:
Ah okay you are right
Nice idea
I would still try to keep the possibility to have a widget with for example an inkwell 
I test your branch and see if I can still add an Inkwell

Feichtmeier:
yaruyaru2234234234

```dart
YaruCarousel(
        children: List.generate(
          length,
          (_) => InkWell(
            borderRadius: BorderRadius.circular(10),
            onTap: () => ScaffoldMessenger.of(context)
                .showSnackBar(SnackBar(content: Text('Hi Paul!'))),
            child: Container(
              decoration: BoxDecoration(
                borderRadius: BorderRadius.circular(10),
                border: Border.all(
                    color: Theme.of(context)
                        .colorScheme
                        .onSurface
                        .withOpacity(0.1)),
                image: DecorationImage(
                    fit: BoxFit.contain,
                    image: AssetImage('assets/ubuntuhero.jpg')),
              ),
            ),
          ),
        ),
        height: 400,
      )
```

Sadly this gesturedetector is ignored for example with an inkwell
Maybe we could add some kind of indicator instead?

Jupi007:
Thanks for pointing me this problem :)
I fixed it with an `IgnorePointer` widget.
Like this, child events are only fired on the current page.

Feichtmeier:
yayayayy

Nice! nvm the indicator idea, this is better

Feichtmeier (approved):

PR #118, opened 2022-05-12
YaruSelectableContainer fix inner border-radius

Jupi007:
The inner border-radius shouldn't use the same value as the outer one.
It needs to be computed with the padding.

**Before:**

Capture d’écran du 2022-05-12 22-46-24

**After:**

Capture d’écran du 2022-05-12 22-46-08


Feichtmeier (approved):
awesome, thanks!

PR #179, opened 2022-08-10
Add new YaruPageItemTitle widget

### LINKED ISSUE #174: YaruPageItem title creates new lines
When becoming too long, the `YaruPageItem` wrap the title text into a new line, which isn't looking good:

Capture d’écran du 2022-08-07 17-29-18

A potential "fix" for apps could be to set the exactly needed sidebar width, but with internationalization it is not possible to deal with so much line size.
IMO the best solution is to ellipse the text like in nautilus:

Capture d’écran du 2022-08-07 17-36-43

As the title is not a `string` but a free widget, I propose to locally change the `max-line` to `1` and the `overflow` to `TextOverflow.ellipsis`, with something like a `YaruPageItemTitle` widget, so this is optional.

Jupi007:
Add a new `YaruPageItemTitle` widget, which alter the current text style with:
- `maxLines`: **1**
- `overflow`: `TextOverflow.ellipsis`

You can pass a child widget, or directly a string with the `YaruPageItemTitle.text` constructor.

**Before:**

Capture d’écran du 2022-08-10 17-52-05

**After:**

Capture d’écran du 2022-08-10 17-51-45

Closes #174

### DIFF
--- a/example/lib/example_page_items.dart
+++ b/example/lib/example_page_items.dart
@@ -23,7 +23,7 @@ const _lorem =
 
 final examplePageItems = <YaruPageItem>[
   YaruPageItem(
-      titleBuilder: (context) => Text('YaruRow'),
+      titleBuilder: (context) => YaruPageItemTitle.text('YaruRow'),
       iconData: YaruIcons.format_unordered_list,
       builder: (_) => YaruPage(children: [RowList()]),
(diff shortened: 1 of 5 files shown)


Feichtmeier (approved):

Jupi007:
I'll do a PR for settings :)

PR #199, opened 2022-09-03
Fix Flutter 3.3 deprecations

Jupi007:
Fix those deprecations from Flutter 3.3:
- info • 'backgroundColor' is deprecated and shouldn't be used. Use colorScheme.background instead. This feature was deprecated after v3.3.0-0.5.pre. • lib/src/pages/layouts/yaru_wide_layout.dart:90:62 • deprecated_member_use
- info • 'backgroundColor' is deprecated and shouldn't be used. Use colorScheme.background instead. This feature was deprecated after v3.3.0-0.5.pre. • lib/src/utilities/yaru_banner.dart:115:41 • deprecated_member_use

### DIFF
--- a/lib/src/pages/layouts/yaru_wide_layout.dart
+++ b/lib/src/pages/layouts/yaru_wide_layout.dart
@@ -87,7 +87,8 @@ class _YaruWideLayoutState extends State<YaruWideLayout> {
                             fontSize: 13,
                             fontWeight: FontWeight.w500,
                           ),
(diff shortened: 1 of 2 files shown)


Feichtmeier (approved):
Thanks! Wow many things changed around our libs with the new versions. Any visual regressions?

Jupi007:
I didn't noticed any problem nor regression with recent version of Flutter :)

PR #204, opened 2022-09-17
Master-detail: use the new page transitions in yaru.dart

jpnurmi:
Vertical [page transitions](https://github.com/ubuntu/yaru.dart/pull/206) in landscape mode vs. horizontal in portrait:

https://user-images.githubusercontent.com/140617/190850908-b2e530ba-4f29-45bd-aa49-9890793dbfbd.mp4


Feichtmeier:
Thank you ☺️

jpnurmi:
Hmm, do we need a way to disable or override the transitions?

Feichtmeier:
Currently not I think this is a very good solution for both small and big windows

Feichtmeier:
https://user-images.githubusercontent.com/15329494/190855735-dabb6afd-e2c8-441c-aaa4-fbe42550635d.mp4

Tested, looking good :+1: 

The dependency on yaru is not perfect, but currently I don't think anyone would use this lib here without yaru

Jupi007:
> The dependency on yaru is not perfect, but currently I don't think anyone would use this lib here without yaru

I had the same thought :)

PR #207, opened 2022-09-18
Fix version constraints

jpnurmi:


pubspec.yaml
```diff
@@ -5,8 +5,8 @@ homepage: https://github.com/ubuntu/yaru_widgets.dart
 version: 1.1.3
 
 environment:
-  sdk: ">=2.17.0 <3.3.2"
-  flutter: ">=3.3.2"
+  sdk: ">=2.17.0 <3.0.0"
```
jpnurmi, line 8: this is the Dart SDK i.e. the language version. anything less than 3.0 is accepted but 3.0 is not because it's going to incompatible

pubspec.yaml
```diff
@@ -5,8 +5,8 @@ homepage: https://github.com/ubuntu/yaru_widgets.dart
...
+  flutter: ">=3.3.0"
```
jpnurmi, line 9: people are strongly encouraged to upgrade to 3.3.2 because of the important bug fixes but technically (API-wise) only 3.3.0 is required

Feichtmeier:
Thank you for the eluding comments :+1: :smile_cat:

PR #209, opened 2022-09-20
Hide control buttons when first/last page reached

### LINKED ISSUE #208: Enhancement for arrows on image carousel
The difference between the active and deactivated arrows in the image carousel are too subtle IMHO. I think a better solution would be if the images could just loop ♾️ or alternatively, to hide the left arrow on the first image (and right arrow for the last ofc).

For accessibility the keyboard left/right arrows should work when I click the arrows, the dots and when the image is enlarged.

image

Jupi007:
[Capture vidéo du 20-09-2022 10:27:41.webm](https://user-images.githubusercontent.com/36476595/191211009-0a3a89c9-a0ca-41b6-a6f3-53be22a769da.webm)

Closes #208

### DIFF
--- a/lib/src/utilities/yaru_carousel.dart
+++ b/lib/src/utilities/yaru_carousel.dart
@@ -177,15 +177,20 @@ class _YaruCarouselState extends State<YaruCarousel> {
     Widget icon,
   ) {
     return Positioned.fill(
-      child: Align(
-        alignment: alignement,
-        child: OutlinedButton(
-          style: OutlinedButton.styleFrom(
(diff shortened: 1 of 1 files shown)


Feichtmeier (approved):
LGTM!

Jupi007:
About the infinite scroll, the `PageView` doesn't allow it.
And it require this kind of trick: https://github.com/jzoom/infinity_page_view/blob/master/package/lib/infinity_page_view.dart

madsrh:
👌 looks good @Jupi007

PR #217, opened 2022-09-23
YaruOptionButton: replace iconData with a child widget

jpnurmi:
Allow passing an arbitrary widget without limiting it to an Icon, and set the desired icon size via IconTheme. This allows passing an SVG or wrapping the icon with a badge, for example.


Feichtmeier (approved):
Thanks. Sorry for doing this iconData stuff everywhere. Makes really no sense :(

jpnurmi:
I might push more of these later. :) Can we hold off with the next release for a little while because these are breaking changes?

Feichtmeier:
> 

Totally! Everything fine for software currently :+1: Thank you for the help!

PR #219, opened 2022-09-23
Implement Yaru(Round)IconButton as IconButton

jpnurmi:
With `IconButton` we get accessibility support, icon theming, focus management, mouse cursor, ... for free.


jpnurmi:
It ensures the same 40x40 size by default but can be now styled to have any dimensions. It has the same API as `IconButton` (except the old color properties that are now available in the style) which is why the proposal to name it `YaruIconButton`.

Feichtmeier:
Sounds good. The only reason why I've used the custom inkwell was because the icon buttons get stretched all the time. If this isn't the case anymore I'm fine with this

PR #225, opened 2022-09-25
Implement YaruColor(Picker)Button as OutlinedButton

jpnurmi:
Based on the same idea as #219. Exposes styling and focus management. Furthermore, proposing to drop the "Picker" word because the button doesn't show a picker but only indicates the given color. What do you think?


Feichtmeier:
Thanks!
Sounds nice!

Eventually you could add the option between a circle shape and this rounded rect?
There is also https://github.com/ubuntu/yaru_widgets.dart/blob/main/lib/src/controls/yaru_color_disk.dart
which can be used to create a nice color picker array/grid

jpnurmi:
Hmm, I wonder if we could merge these two!

Feichtmeier:
yeah : D it's almost the same idea
The only difference is that you can set the color disk to be selected with the primary color as a border

jpnurmi:
Hmm, I think this was superseded by #226.

PR #228, opened 2022-09-27
Extract page controller from carousel

Jupi007:
This allow to control it from anywhere.
We keep the same functionnalities as before.


Feichtmeier (approved):
Very cool - Thank you!

PR #232, opened 2022-10-05
YaruRow: make the description a widget

jpnurmi:


lib/src/pages/rows/yaru_row.dart
```diff
@@ -103,13 +103,13 @@ class YaruRow extends StatelessWidget {
                     if (description != null)
                       Padding(
                         padding: const EdgeInsets.only(top: 4.0),
-                        child: Text(
-                          description!,
+                        child: DefaultTextStyle(
                           style: enabled
-                              ? Theme.of(context).textTheme.bodySmall
-                              : Theme.of(context).textTheme.bodySmall?.copyWith(
+                              ? Theme.of(context).textTheme.bodySmall!
```
jpnurmi, line 108: `DefaultTextStyle` requires a non-nullable `TextStyle`. can we assume that `bodySmall` is set?

jpnurmi: `ListTile` seems to do a similar assumption: https://github.com/flutter/flutter/blob/69fc4fb622384d3858d6066afc8b81370a3e7f23/packages/flutter/lib/src/material/list_tile.dart#L649-L659

Feichtmeier (approved):
😎👍 much better
No idea which nonsense was in my head at this time

PR #234, opened 2022-10-05
Yaru*Row: remove the width property

jpnurmi:
ListTile doesn't have such property either. In the settings app, there are only a handful of cases where the width property is used. Often it can be set for the entire section/column/list instead of each individual tile.


Feichtmeier (approved):
Absolutely!

PR #240, opened 2022-10-06
Analysis improvements and fixes

Feichtmeier:


Feichtmeier:
Uh nice ty!

Was all easy to fix except:

prefer_null_aware_method_calls: true

for:
```dart
if (widget.onChange != null) {
  widget.onChange!(_isExpanded);
}
```

with 

```dart
final void Function(
    bool isExpanded,
  )?
```

jpnurmi:
```dart
widget.onChange?.call(_isExpanded);
```

jpnurmi:
I don't know why dart allows
```dart
widget.onChange!(_isExpanded);
```
but not
```dart
widget.onChange?(_isExpanded);
```
we have to add an explicit `.call()`...

Feichtmeier:
Thanks! Was not aware of this way of calling methods!

PR #243, opened 2022-10-06
Add Tooltips (and so Semantics) to layouts navigation

Jupi007:
Capture d’écran du 2022-10-06 20-17-00
Capture d’écran du 2022-10-06 20-17-16

This is useful when labels are elided (and it also add Semantics data);

### DIFF
--- a/example/lib/example_page_items.dart
+++ b/example/lib/example_page_items.dart
@@ -15,66 +15,65 @@ import 'pages/tabbed_page_page.dart';
 import 'pages/tile_page.dart';
 
 final examplePageItems = <YaruPageItem>[
-  YaruPageItem(
-    titleBuilder: (context) => YaruPageItemTitle.text('YaruBanner'),
+  YaruPageItem.titleFromLabel(
+    label: 'YaruBanner',
     builder: (context) => const BannerPage(),
     iconBuilder: (context, selected) => selected
         ? const Icon(YaruIcons.image_filled)
         : const Icon(YaruIcons.image),
   ),
-  YaruPageItem(
-    titleBuilder: (context) => YaruPageItemTitle.text('YaruCarousel'),
+  YaruPageItem.titleFromLabel(
+    label: 'YaruCarousel',
     builder: (_) => const CarouselPage(),
     iconBuilder: (context, selected) => const Icon(YaruIcons.refresh),
   ),
-  YaruPageItem(
-    titleBuilder: (context) => YaruPageItemTitle.text('YaruColorDisk'),
+  YaruPageItem.titleFromLabel(
+    label: 'YaruColorDisk',
     builder: (context) => const ColorDiskPage(),
     iconBuilder: (context, selected) => const Icon(YaruIcons.color_select),
   ),
-  YaruPageItem(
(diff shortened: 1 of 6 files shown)


jpnurmi:
Could we consider replacing `iconBuilder` and `titleBuilder` with a `tileBuilder` and offer the recommended tile widget as a separate building block? This way, users could wrap it with anything or even replace the whole thing with something else.

Before:
```dart
YaruPageItem(
  iconBuilder: (_) => Icon(Icons.foo),
  titleBuilder: (_) => Text('bar'),
)
```

After:
```dart
YaruPageItem(
  tileBuilder: (_) => YaruPageTile(
    icon: Icon(Icons.foo),
    title: Text('bar'),
  ),
)
```

Feichtmeier:
Sounds good - but how would we do the selection index then?

jpnurmi:
Could we handle clicks below so that it works automatically as long as the supplied tile is non-interactive like `Yaru(Page)Tile` is? If you'd supply an interactive tile that blocks clicks, you'd have to rebuild the page yourself (`YaruMasterDetailPage` constructor should take the current index). Would that work?

Feichtmeier:
I am unable to imagine this currently 😄 
I think if Paul agrees and you are sure that it would work paul cancels this PR and you create a new one?

jpnurmi:
I'm 110% _unsure_ if it works. :rofl: I'm just brainstorming ideas to make it more open because we would like more control over the tile layout and contents in the firmware updater.

Feichtmeier:
Currently both layouts build their tiles with the title and icon information
I don't know how we would do this when we forward our own tile
Eventually we could rather forward more builders like subtitle widget, tooltip string and so on
Then both layouts can keep the way they build their tiles

Jupi007:
Imo, it is a really good idea :)
But if we go in this way, I don't know if we'll still be able to switch between layouts as simple as we do in the example.

Maybe still merge this PR?
_After I fix those conflicts 😅_

jpnurmi:
> I don't know if we'll still be able to switch between layouts as simple as we do in the example.

True, it's a compromise between convenience vs. flexibility. :)

Even though, is it actually a common use case to switch between the master-detail and navi-rail layouts in a real app? Perhaps the two could be even different examples because the "Layout" item doesn't quite seem to fit with the other YaruXxx items. :)

Feichtmeier:
Yes, the MD page is for very long lists
CL is for 3-10 I would say

jpnurmi:
I hadn't actually noticed that the example had been changed to present the same page items in both layout modes. Nice! :wink:

Jupi007:
About this PR, I think we should still add a Tooltip at least in `YaruNavigationRailItem`.
As labels can be hidden in this layout, it is imo important.
It can be also useful in `YaruMasterTile`.

jpnurmi:
Sorry, I missed this. Sure, we could add a tooltip property at least to `YaruNavigationRailItem`! :+1:

`YaruMasterTile` can have multiple children so it's probably better to let users wrap the title & subtitle with separate Tooltip widgets.

image

Jupi007:
I close this PR to redo another one.

PR #244, opened 2022-10-06
Remove YaruPage in favor of the padding etc. constants

jpnurmi:
`YaruPage` is not as generic as it sounds because it wraps `SingleChildScrollView` + `Column` which is problematic if you want to put any scrollable content such as `ListView` or `GridView` in it. The goal of `YaruPage` is to provide appropriate padding, which can be more effectively set via the `padding` property of any scrollable widget.


jpnurmi:
### Bad
```dart
SingleChildScrollView(
  child: Padding(
    padding: ...
    child: ListView(
      shrinkWrap: true, // :(
      ...
    ),
  )
)
```

### Good
```dart
ListView(
  padding: ...
  ...
)
```

Feichtmeier:
Yeah it's really unpredictable what this does to performance compared to list and gridviews!
+1

PR #247, opened 2022-10-07
YaruMasterDetailPage: allow managing the page index

jpnurmi:
Technically, you could even rebuild the page if you wanted:
```dart
YaruMasterDetailPage(
  initialIndex: _idx,
  onSelected: (index) => setState(() => _idx = index),
)
```
Even though the real motivation is to be able to specify a page on startup _and_ that this might open up the possibility to allow full tile builders (#243).


Feichtmeier:
Neat!
Could you do the same for the compact layout? So they both stay close to each other

jpnurmi:
It already had `initialIndex` so I just added the callback: https://github.com/ubuntu/yaru_widgets.dart/pull/249

PR #248, opened 2022-10-07
MasterDetailPage: replace "page item" with indexed builders

jpnurmi:
One app might have a list of page items whereas the other has something totally different, such as a list of devices. Dropping the page item concept allows apps to operate directly using their own data structures without having to map them into a list of page items.


jpnurmi:
Potential next steps:
- remove the app bar as discussed earlier (kills a dependency to `titleBuilder`)
- replace `iconBuilder` & `titleBuilder` by `tileBuilder`

PR #253, opened 2022-10-09
YaruPortraitLayout: remove the FAB

jpnurmi:
Let apps decide where they want to put the back button if they are not using an app bar.

image


Feichtmeier:
Lol 😆 where there hell came this from

Issue #257, opened 2022-10-09
Round icon button is now a bar


Feichtmeier:
![Bildschirmfoto 2022-10-09 um 17 03 48](https://user-images.githubusercontent.com/15329494/194764167-a5885de3-3216-4d70-8aa2-515550a04180.png)

jpnurmi:
Btw, is `YaruRoundToggleButton` not the same as `IconButton` with `isSelected` property?

jpnurmi:
`YaruRoundToggleButton` vs. `YaruIconButton(isSelected: ...)` are visualized slightly different.

Feichtmeier:
How bad is the difference?

jpnurmi:
`YaruRoundToggleButton` looks arguably better because it has a) background highlight in selected mode and b) the highlight & splash are not using the primary color.

| | `YaruRoundToggleButton` | `YaruIconButton` |
|---|---|---|
| unselected | ![image](https://user-images.githubusercontent.com/140617/194768244-7092aac0-5f15-46bd-b704-e77f0fbc5c4b.png) | ![image](https://user-images.githubusercontent.com/140617/194768632-6897c372-9946-4081-9bf6-d736d8958097.png) |
| unselected, hovered | ![image](https://user-images.githubusercontent.com/140617/194768210-54c0ccc2-da8c-49ff-ac9c-d58a11696786.png) | ![image](https://user-images.githubusercontent.com/140617/194768676-1f37638b-a9d6-4e67-b2cd-50a1da2041b3.png) |
| unselected, pressed | ![image](https://user-images.githubusercontent.com/140617/194768369-34affaea-981f-49be-b213-ebdd48d3b3af.png) | ![image](https://user-images.githubusercontent.com/140617/194768702-ed610816-53bb-4944-b305-40c53b9343db.png) |
| selected | ![image](https://user-images.githubusercontent.com/140617/194768421-5b556944-e156-4f38-b4c6-a8c82d1ef06a.png) | ![image](https://user-images.githubusercontent.com/140617/194768735-b727d703-6e31-4db9-a737-c969747a77a6.png) |
| selected, hovered | ![image](https://user-images.githubusercontent.com/140617/194768446-0445705d-ba8e-451c-8a3e-abb13bf71036.png) | ![image](https://user-images.githubusercontent.com/140617/194768759-93a168d1-e168-4191-a585-0290ae1cc8e7.png) |
[...]

PR #258, opened 2022-10-09
Example: center the toggle button to avoid stretching it

jpnurmi:
| Before | After |
|---|---|
| Screenshot from 2022-10-09 17-55-01 | Screenshot from 2022-10-09 17-54-55 |

Fixes: #257


Feichtmeier:
Ty!

Issue #255, opened 2022-10-09
YaruMasterDetailPage wish-list


jpnurmi:
- [x] full control over master list tiles
  - replace "icon" & "title" with a "tile" builder?
  - tooltips, semantics, badges, animations, custom tiles, ...
- [x] full control over detail page app bar
  - replace "title" with an "app bar" builder?
  - leading icon, trailing actions, bottom widget, ...
- [x] themable constraints, padding, spacings, ...

**TODO**: update the docs when all the refactoring is done!

Feichtmeier:
> replace "title" with an "app bar" builder?

I don't think we need this if each page is itself either a column (appbar/page) or a scaffold (appbar, body) :)

jpnurmi:
i'm afraid i screwed up that possibility with the animations... :fearful: 

[appbar-vs-page.webm](https://user-images.githubusercontent.com/140617/194766244-ac5f73e4-0bbc-4459-b475-58c049f26a58.webm)

Feichtmeier:
I dont understand

jpnurmi:
Bad screencast quality. :slightly_frowning_face: The problem is that the app bar gets included in the page transitions if it's inside the content.

Feichtmeier:
Hm but what about pages that make use of model classes in the app bar?
I think it would be better to include them on the page directly in a scaffold

Otherwise you need to suck in the model inside the appBarBuilder

Widget appBarBuilder(BuildContext context) {
final model = context.watch<MyModel>();
return AppBar(flexibleSpace: TextField(onChanged: ()=> model.bla = .... ));
}
and so on

jpnurmi:
It's a bit hackish but one possibility to exclude `AppBar` from the page transition is to wrap it with a `Hero` widget. What makes it a bit inconvenient is that `Scaffold.appBar` requires a `PreferredSizeWidget` so you can't pass `Hero` directly...

```dart
Scaffold(
  appBar: PreferredSize(
    preferredSize: Size(
      0,
      Theme.of(context).appBarTheme.toolbarHeight ?? kToolbarHeight,
    ),
    child: Hero(
      tag: 'no-animation',
      child: AppBar(
        title: const Text('Title'),
      ),
    ),
  ),
  body: const MyPage(),
)
```

Feichtmeier:
IMO that the app bar is included in the transition is not bad
It's part of the page

jpnurmi:
[...]

PR #260, opened 2022-10-10
Add YaruBackButton

jpnurmi:
This is a pre-requisite for the master-detail app bar removal (#255), also discussed in ubuntu-flutter-community/software#354.

P.S. Proposing `YaruBackButton` instead of `YaruIconButton.back` because the former is easier to discover :slightly_smiling_face:

**NOTE**: rectangular vs. circular shape change and a dependency to **yaru_icons** - are these acceptable?

| Before | ~~After~~ | After
|---|---|---|
| Screenshot from 2022-10-10 11-45-08 | Screenshot from 2022-10-10 11-44-55 | Screenshot from 2022-10-10 11-45-08 |

### DIFF
--- a/example/lib/main.dart
+++ b/example/lib/main.dart
@@ -63,7 +63,6 @@ class _HomeState extends State<Home> {
           ? _CompactPage(configItem: configItem)
           : YaruMasterDetailPage(
               leftPaneWidth: 280,
-              previousIconData: YaruIcons.go_previous,
               length: pageItems.length,
               tileBuilder: (context, index, selected) => YaruMasterTile(
                 leading: pageItems[index].iconBuilder(context, selected),
--- a/lib/src/controls/yaru_back_button.dart
+++ b/lib/src/controls/yaru_back_button.dart
@@ -1,0 +1,31 @@
+import 'package:flutter/material.dart';
+import 'package:yaru_icons/yaru_icons.dart';
+
+import 'yaru_icon_button.dart';
+
+/// A Yaru style back button.
+class YaruBackButton extends StatelessWidget {
+  /// Creates a [YaruIconButton].
+  const YaruBackButton({super.key, this.onPressed});
+
+  /// An optional callback that is called when the button is pressed.
+  ///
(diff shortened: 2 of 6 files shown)


jpnurmi:
Plan:
1. add `YaruBackButton`
2. add `YaruDetailPage` with a [hero](https://github.com/ubuntu/yaru_widgets.dart/issues/255#issuecomment-1272597704) for page transitions and Yaru-style back button
3. remove the app bar in `YaruLandscapeLayout`

jpnurmi:
Thanks, I changed it back to a rectangular shape.

Feichtmeier:
I must admit that I do not like to pull in yaru icons as a dep
I do not even like yaru.dart as a dep : D BUt that's in
Is it really needed to include the yaru icons dep?

jpnurmi:
Well, it kinda defeats the purpose of `YaruBackButton` one always has to pass an icon to get the correct looks. :slightly_smiling_face: And if `YaruBackButton` needs to be populated with an icon, then `YaruMasterDetailPage` and `YaruDetailPage` need to be populated with back icons, too...

Feichtmeier:
> Well, it kinda defeats the purpose of `YaruBackButton` one always has to pass an icon to get the correct looks. slightly_smiling_face And if `YaruBackButton` needs to be populated with an icon, then `YaruMasterDetailPage` and `YaruDetailPage` need to be populated with back icons, too...

Fine but I think then we need to update the readme, because this lib stopped being an addition to material.dart, rather an addition to yaru.dart and yaru_icons.dart :smile_cat: 
And eventually we could then in some other change remove all other IconData/icon parameters (I remember this is in the dialogs)

jpnurmi:
Indeed, we'll have to review the docs as soon as things calm down 👌

jpnurmi:
> this lib stopped being an addition to material.dart, rather an addition to yaru.dart and yaru_icons.dart 😸

Sounds fine considering the name of the package ☺️

Feichtmeier:
Okay :+1: LGTM merge this and the other open PRs in the order you think you won't get any conflicts :P

PR #262, opened 2022-10-10
Use widgets as parameters in Yaru[Section,TabbedPage,Banner]

d-loose:
I've replaced `String` and `IconData` parameters with `Widget` and `Icon`, respectively.
The `summary` parameter in `YaruBanner` was redundant, so I removed it.

### DIFF
--- a/lib/src/pages/yaru_tabbed_page.dart
+++ b/lib/src/pages/yaru_tabbed_page.dart
@@ -21,8 +21,8 @@ class YaruTabbedPage extends StatefulWidget {
     this.onTap,
   });
 
-  /// A list of [IconData] used inside the tabs - must have the same length as [tabTitles] and [views].
-  final List<IconData> tabIcons;
+  /// A list of [Icon]s used inside the tabs - must have the same length as [tabTitles] and [views].
+  final List<Icon> tabIcons;
 
   /// The list of titles as [String]s - must have the same length as [tabIcons] and [views].
   final List<String> tabTitles;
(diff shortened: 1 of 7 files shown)


Feichtmeier:
Thans @d-loose 
The summary parameter is used inside software, so please either keep it or replace it with a Widget as well :)

For example
https://github.com/ubuntu-flutter-community/software/blob/edadfb45cc271ce88832a5254ae1bc1afe418c90/lib/store_app/explore/start_page.dart#L213-L221

Feichtmeier:
Oh you got the subtitle widget! Sorry :) LGTM then

Feichtmeier:
however, could you rename it to "subtitle" then, no idea why I cam up with this stupid name? I feel like the Widget suffix is not needed

lib/src/pages/yaru_tabbed_page.dart
```diff
@@ -21,8 +21,8 @@ class YaruTabbedPage extends StatefulWidget {
     this.onTap,
   });
 
-  /// A list of [IconData] used inside the tabs - must have the same length as [tabTitles] and [views].
-  final List<IconData> tabIcons;
+  /// A list of [Icon]s used inside the tabs - must have the same length as [tabTitles] and [views].
+  final List<Icon> tabIcons;
```
Jupi007, line 25: Here I think the type should be List<Widget>
Example: animated icons are widgets.

Jupi007 (changes requested):

lib/src/pages/yaru_tabbed_page.dart
d-loose: Ah yes, that makes a lot of sense - thanks!

Jupi007 (approved):
LGFM thanks 😊

Feichtmeier:
I think the dark style for banner names was changed by the change of the modified text to the name widget
At least it's an issue in the web build since this commit

d-loose:
Interesting - I can't reproduce it locally, works fine when I build it. I'll look into it tomorrow!

PR #264, opened 2022-10-10
unify YaruIconButton and YaruRoundToggleButton

d-loose:
An attempt to simplify and unify the icon and toggle buttons.
Instead of having two different widgets, `YaruIconButton` now behaves like `YaruRoundToggleButton` if `isSelected` is set.

Is there an easy way of obtaining the overlay color from existing styles rather than hardcoding it?


Feichtmeier:
Thank you @d-loose ! :+1: 

> 

what do you mean with overlay color? this? colors.onSurface.withOpacity(0.8); then just keep it like this please, it is easy to handle and I know what it means :smile: 

Could you add before/after screenshots for light and dark ?

d-loose:
@Feichtmeier concerning the overlay color I meant [this part](https://github.com/ubuntu/yaru_widgets.dart/pull/264/files#diff-bc4b299e678e787c164af711f0c6b6007c8b368397fe66793e66d875dffce9c2R58-R80), which I basically stole from [here](https://github.com/flutter/flutter/blob/master/packages/flutter/lib/src/material/icon_button.dart#L903).

Feichtmeier:
looks good thank you!

PR #266, opened 2022-10-10
YaruLandscapeLayout: allow left pane resizing

Jupi007:
- [x] Add min-width ;
- [x] save pane width when switching between layouts (landscape, portrait) ;
- [x] add an option to disable this behavior ;
- [x] fix infinite pane resizing, and window resizing.

[Capture vidéo du 10-10-2022 19:22:14.webm](https://user-images.githubusercontent.com/36476595/194921347-68bdb88b-f9f4-4a03-97ca-807982046f44.webm)


Jupi007:
Thanks :)

Well, that's a good point, but the `YaruMasterDetailPage` isn't supporting well rtl by default. Ex: the separator isn't placed correctly: it is on right instead of the left.
If you want, I'll do another PR to fix all that (and maybe also the compact layout), and I'll rename the left pane to something like `sidebar`

Jupi007:
Sorry, this is wip :)

Jupi007:
Final version:

[Capture vidéo du 10-10-2022 21:51:33.webm](https://user-images.githubusercontent.com/36476595/194942445-f9d19cfe-cff1-445b-ac3a-3c18599f9399.webm)

lib/src/pages/layouts/yaru_landscape_layout.dart
```diff
@@ -47,13 +63,40 @@ class YaruLandscapeLayout extends StatefulWidget {
...
+  @override
+  void didUpdateWidget(covariant YaruLandscapeLayout oldWidget) {
+    super.didUpdateWidget(oldWidget);
+
+    final width = MediaQuery.of(context).size.width;
+
+    // Avoid left pane to overflow when resizing the window
+    if (widget.leftPaneResizing &&
+        _leftPaneWidth >= width - widget.pageMinWidth) {
+      setState(() {
```
jpnurmi, line 95: `setState` could be removed because the framework is just about to call `build` after `didUpdateWidget`

Feichtmeier (approved):
Neat! Thank you

lib/src/pages/layouts/yaru_master_detail_page.dart
```diff
@@ -32,6 +34,9 @@ class YaruMasterDetailPage extends StatefulWidget {
     required this.titleBuilder,
     required this.pageBuilder,
     required this.leftPaneWidth,
+    this.leftPaneResizing = true,
```
jpnurmi, line 37: nit-pick: this sounds like the pane is currently resizing :)

lib/src/pages/layouts/yaru_landscape_layout.dart
```diff
@@ -63,51 +106,136 @@ class _YaruLandscapeLayoutState extends State<YaruLandscapeLayout> {
...
+            onPanStart: (details) => setState(() {
+              _isDragging = true;
+              _initialPaneWidth = _leftPaneWidth;
+            }),
+            onPanUpdate: (details) => setState(() {
+              _paneWidthMove += details.delta.dx;
+              final width = _initialPaneWidth + _paneWidthMove;
+
+              final previousPaneWidth = _leftPaneWidth;
+              final contextWidth = MediaQuery.of(context).size.width;
```
jpnurmi, line 214: this assumes that the layout covers the whole window?

lib/src/pages/layouts/yaru_master_detail_page.dart
```diff
@@ -32,6 +34,9 @@ class YaruMasterDetailPage extends StatefulWidget {
...
+    this.leftPaneMinWidth = 175.0,
+    this.pageMinWidth = 310.0,
```
jpnurmi, line 39: hmm, these are a bit arbitrary numbers :)

lib/src/pages/layouts/yaru_landscape_layout.dart
jpnurmi: maybe LayoutBuilder could be used to get the actual available width?

lib/src/pages/layouts/yaru_landscape_layout.dart
Jupi007: Indeed you're right.

lib/src/pages/layouts/yaru_master_detail_page.dart
Jupi007: Hmm, what about `allowLeftPaneResize`?

jpnurmi: :+1: I'm not a native English speaker either but it sounds better in my head :slightly_smiling_face:

lib/src/pages/layouts/yaru_master_detail_page.dart
Jupi007: Indeed, but we need ones :)

At 175 width, the left pane only show a few letters, 150 is imo too few.
And 310, is the half of the width (620) where the layout switch to portrait layout.

But I'm open to some other values or other way do add defaults :)

lib/src/pages/layouts/yaru_landscape_layout.dart
Jupi007: Oh, I didn't know this.

lib/src/pages/layouts/yaru_master_detail_page.dart
jpnurmi: Depends also on the system font. What about moving that `620` to `constants.dart` and using `kFoo / 2`?

Jupi007: The `175` is totally arbitrary, I admit I set this value just because I had feel it was the best one.
But honestly, this value is kind of edge use case. It looks good "by default", when an user want to collapse the sidebar at maximum. This is just prevent any widget to overflow and broke the layout. If the app is define a tile which can't be collapsed so much, he can overwrite this value.
At least it is how I see this value :)

About the `620` one, I see you created a `YaruMasterDetailThemeData` with this value hardcoded as fallback.
So, should we use the same constant there?

jpnurmi: sharing the same constant would help to make the intention/origin clear. when i saw 310 at first, i didn't realize it was the half. 😁

jpnurmi: Btw, I wrote `kFoo` because I have no idea what the constant should be called. Do you have something in mind?

Jupi007: What about something like simply `kMasterDetailBreakpoint`?

jpnurmi: Sure, I prefixed them all with `kYaru...` to distinguish them from Flutter's constants, though.

Jupi007: Okay, so let's do `kYaruMasterDetailBreakpoint` :)

Jupi007:
Thanks a lot for your review @jpnurmi 🙂
It is fine now?

jpnurmi (commented):
I think it's fine even though I'm still mentally struggling with the 175 magic number. :)

Is it based on the example app? Notice that the user might have the "large text" accessibility setting toggled on in which case the tiles need significantly more space to look sensible. On the other hand, an app might have tiles without icons and then less would be sufficient. I think it's super hard even for the user that knows what kind of tiles they have, to calculate any meaningful value.

Jupi007:
> Is it based on the example app? Notice that the user might have the "large text" accessibility setting toggled on in which case the tiles need significantly more space to look sensible.

Yes, it is based on the example, which is the default look.
About larger text cause to accessibility, it is imo not a big problem: if the user collapse the sidebar, it is to have more space for the page, so I don't think he would want to read the labels in this situation :)
This value is only here to avoid layout problem.

> On the other hand, an app might have tiles without icons and then less would be sufficient. I think it's super hard even for the user that knows what kind of tiles they have, to calculate any meaningful value.

In that case, if the app change the tiles to something smaller, this value can be customized in the constructor. Btw it is the same for any larger tile, which will require a larger min width.
This minWidth default value is only here to have something good looking in the default configuration and using default tile :)

jpnurmi:
What about grouping anything resizing-related as required properties into a "constraints" or "delegate" type of object? It's arguably less convenient for users but forcing them to think about the size constraints prevents such surprises that may happen if we allow them to just set one property and they expect it to magically do the right thing (whatever it means in their case).

```dart
YaruMasterDetailPage(
  resizeDelegate: YaruMasterResizeDelegate( // or "Constraints"?
    minPaneWidth: 123,
    minPageWidth: 456,
  ),
),
```

jpnurmi:
Or taking it a bit further to earn the "delegate" name ;)
```dart
abstract class YaruMasterSizeDelegate {
  const YaruMasterSizeDelegate();

  double calculatePaneWidth(double availableWidth);
  double calculatePageWidth(double availableWidth);
}

class YaruMasterFixedDelegate implements YaruMasterSizeDelegate {
  const YaruMasterFixedDelegate({required this.paneWidth});

  final double paneWidth;

  @override
  double calculatePaneWidth(double availableWidth) {
    return paneWidth;
  }

  @override
  double calculatePageWidth(double availableWidth) {
    return availableWidth - paneWidth;
  }
}


class YaruMasterResizeDelegate extends YaruMasterSizeDelegate {
  const YaruMasterResizeDelegate({
    required this.preferredPaneWidth,
    required this.minPaneWidth,
    required this.maxPaneWidth,
  });

  final double preferredPaneWidth;
  final double minPaneWidth;
  final double maxPaneWidth;

  @override
  double calculatePaneWidth(double availableWidth) {
    return ...;
  }

  @override
  double calculatePageWidth(double availableWidth) {
    return ...;
  }
}
```

Jupi007:
This looks like a good solution :)
Just a note, there is no max pane width, but more a min page width, which makes the max pane width "dynamic".

Do you want to implement this after this PR?

lib/src/pages/layouts/yaru_landscape_layout.dart
```diff
@@ -47,13 +63,38 @@ class YaruLandscapeLayout extends StatefulWidget {
...
+    super.didUpdateWidget(oldWidget);
+
+    final width = MediaQuery.of(context).size.width;
+
+    // Avoid left pane to overflow when resizing the window
+    if (widget.allowLeftPaneResize &&
+        _leftPaneWidth >= width - widget.pageMinWidth) {
+      _leftPaneWidth = width - widget.pageMinWidth;
+      widget.onLeftPaneWidthChange?.call(_leftPaneWidth);
+    }
```
jpnurmi, line 97: there's still an assumption that the widget covers the entire window. shouldn't this check be inside the layout builder instead of relying on media query?

jpnurmi (commented):
We can continue on the vague delegate ideas in separate PRs :)

lib/src/pages/layouts/yaru_landscape_layout.dart
Jupi007: You're right, I had forgotten this one :)

lib/src/pages/layouts/yaru_landscape_layout.dart
```diff
@@ -63,51 +104,141 @@ class _YaruLandscapeLayoutState extends State<YaruLandscapeLayout> {
...
+            }),
+            onPanEnd: (details) => setState(() {
+              _isDragging = false;
+              _paneWidthMove = 0.0;
+            }),
           ),
         ),
-      ],
+      ),
+      height: MediaQuery.of(context).size.height,
```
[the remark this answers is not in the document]
Jupi007: `Positioned.fill` is that I was searching, thanks :)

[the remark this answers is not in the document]
Jupi007: I found another solution, because `Positioned.fill` doesn't have width/height:

```dart
Positioned(
  width: _kLeftPaneResizingRegionWidth,
  top: 0,
  left: 0,
  bottom: 0,
)
```

I don't know why I did though about it before 😅

jpnurmi (approved):
Thanks, LGTM :+1:

Issue #271, opened 2022-10-10
YaruPageItemTitle


jpnurmi:
If a generic Yaru control sets the default text style with help of `DefaultTextStyle`, the application can override the desired style attributes by passing a `TextStyle` instance to the `Text` widget it creates. `YaruPageItemTitle` might not be needed?

Feichtmeier:
I have no idea this is a question for @Jupi007

Issue #259, opened 2022-10-10
YaruCompactLayout: wrong label wrap


Jupi007:
Oh, this is annoying...
I didn't though about a small word followed by a long one (which is quite common in French).

The only solution I found would be:

```dart
var label = 'Mes applications';
// ...
label.replaceAll(' ', '\u00A0');
```

![Capture d’écran du 2022-10-10 10-48-24](https://user-images.githubusercontent.com/36476595/194833821-1ba44562-cdfe-4be5-b8dc-3f2f0b5d97b6.png)

Which replace all spaces with an unbreakable character.
In fact we cannot expand the text widget as we are in a column, so we need to use a trick...

We could add a patch here, but this is not the good solution.
We need to fix that in yaru_widgets directly.

_Originally posted by @Jupi007 in https://github.com/ubuntu-flutter-community/software/issues/372#issuecomment-1273031090_

jpnurmi:
Sorry, I don't understand. :) How should it look like in this case? If this is language-dependent, shouldn't it be the translators job to decide which whitespace characters are breakable or not?

Jupi007:
Well, if we take the `Mes applications` label example, the ellipsis is done in the white space, which looks like: `mes...` (`my...` in english).
Like this the label makes no sense for the user. I would prefer to have the ellipsis as far as possible in the label, like in my upper screenshot.
But this is not achievable with a text style property, the only way I found is to replace white spaces with unbreakable ones, so the label string becomes a all in one long word.

PR #273, opened 2022-10-11
Remove YaruPageItemTitle

jpnurmi:
The type of the immediate child should not matter. It could be wrapped into a non-Text widget. Just wrap with DefaultTextStyle and let any descendant Text inherit the style. If desired, users can override the defaults in builders.

> Many Widgets take a `child`. Widgets should be entirely agnostic about the type of that child. Don’t use `is` or similar checks to act differently based on the type of the child.

https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo#avoid-interleaving-multiple-concepts-together

Fixes: #271


Feichtmeier:
Did this change anything with the compact layout? I think this is the reason @Jupi007 introduced this, to have the text styled correctly there

Jupi007:
Looking at the code, the text styling is still applied in both compact and master details layouts, so it is fine. It is just now forced.

In fact, I was thinking to make those defauts styles optional, because they are very specifics, especially in the compact layout.
If you want to replace the tile, maybe you don't want to overwrite again the text styling.
But maybe I was wrong, I don't know :)

jpnurmi:
It's not forced but made the default. :) Do you want to add it to the theme? (still missing for CL)

Feichtmeier:
So wrapping it yet in another default text style will overwrite the one beneath? Then it sounds fine

jpnurmi:
`DefaultTextStyle` is great for propagating the desired defaults down the tree when not creating the `Text` yourself. In the builder, one can pass `maxLines`/`overflow`/`style` directly to `Text` constructor.

jpnurmi:
Let's put this on hold until we have made the Compact Layout themable.

Jupi007:
About the `is` thing (I want to justify myself a little 😇):
We first only had the master-details layout, and so I putted the `DefaultTextStyle` inside an optional but recommended `YaruPageItemTitle` widget. Then comes the compact layout, with different style requirements, but using the same data model classes as master-details (and so with `YaruPageItemTitle`). As we wanted to have the possibility to switch between both layouts, I only see the solution of moving the `DefaultTextStyle` inside the layouts themself, but make it conditionally to the `YaruPageItemTitle` widget when used as label.
In retrospect, all this is a bit kind of non sense 😅

Btw, talking about the compact layout, we have this problem: #259
The only solution I found is:
```dart
var label = 'Mes applications';
// ...
label.replaceAll(' ', '\u00A0');
```
But makes this generic would require a widget which would wrap a `Text` one... which isn't pretty at all.
Unless what, apps will need to do the job by themself, but I don't think it is the solution either.

jpnurmi:
While we could have something like `TextStyle titleStyle` in `YaruMasterDetailTheme` and `YaruCompactLayoutTheme`, I feel like the "item" and "tile" builders are already powerful enough for overriding the default text style. Also, having a "title style" on the page theme level kinda goes against the idea of having arbitrary "item" and "tile" builders that may not necessarily even contain anything called "title". What do you think?

Jupi007 (approved):
I think you're right :)
LGFM so 👍

PR #283, opened 2022-10-11
YaruLandscapeLayout: use DividerTheme

Jupi007:
Use DividerTheme instead of hardcoded color to be aligned with the compact layout.

No visual change.

### DIFF
--- a/lib/src/pages/layouts/yaru_landscape_layout.dart
+++ b/lib/src/pages/layouts/yaru_landscape_layout.dart
@@ -123,8 +123,6 @@ class _YaruLandscapeLayoutState extends State<YaruLandscapeLayout> {
     );
   }
 
-  Color get _separatorColor => Colors.black.withOpacity(0.1);
-
   Widget _maybeBuildGlobalMouseRegion(Widget child) {
     if (widget.allowLeftPaneResize) {
       return MouseRegion(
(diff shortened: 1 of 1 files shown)


Feichtmeier:
grafik

Feichtmeier:
better put the screenshots in the PR just to double check next time ☺️ 🙏

Jupi007:
Oh weird, sorry....
As the compact layout was using the same values, I was thinking it was fine 🤕

Jupi007:
Shame on me, the compact layout doesn't use the same separator at all 😵

I have this look:

Capture d’écran du 2022-10-11 17-06-44

And the compact layout looks like:

Capture d’écran du 2022-10-11 17-07-12

Imo, we should unify all that by using `AppBarTheme.shadowColor`, which is the color used in the appbar separator.
Wdyt @Feichtmeier ?

Feichtmeier:
In personally prefer the brighter one

Jupi007:
Only for the compact mode?
Or do you want a screenshot with bright separators in master-details?

Feichtmeier:
For the compact layout in any case
But for MD I'm not sure. Eventually if we would also make the appbar borders bright in the yaru.dart theme.

Jupi007:
Brighter borders would looks like this:

Capture d’écran du 2022-10-11 21-33-45

Feichtmeier:
I can't really say what is better. The bright borders are more consequent but I am not sure about the bigger picture. 🤷

Jupi007:
Honestly, I really prefer with bright borders.

Feichtmeier:
> Honestly, I really prefer with bright borders.

Okay we could test it
I can do the yaru.dart change later today

Jupi007:
All right :)
I'll the PR here.

PR #286, opened 2022-10-12
DialogTitle: don't hide 🫥 properties with strings

### LINKED ISSUE #277: DialogTitle: don't hide 🫥 properties with strings
https://github.com/ubuntu/yaru_widgets.dart/blob/main/lib/src/dialogs/yaru_dialog_title.dart#L18

Again....

Feichtmeier:
@jpnurmi 

Instead of forwarding a function I now have a forced bool parameter so it can be combined with showDialog barrierDismissible

WDYT?

Also
Fixes #277

[dialogs.webm](https://user-images.githubusercontent.com/15329494/195333859-af63228e-54d9-4aa7-898d-b7b5a5d65b37.webm)


jpnurmi:
Perhaps we could expose `YaruCloseButton` as a separate standalone building block just like we have `YaruBackButton`? :)

I like the way `AppBar` works. It offers an easy way to lay out the title with leading and trailing actions, and it does the right thing most of the time i.e. it automatically creates the back button for me if appropriate. Yet, it allows me to replace it with something else when more flexibility is needed.

I'd love `YaruDialogTitle` to be more like `AppBar`. It can create the close button by default but perhaps it's enough to allow replacing it when more customization is needed? Another thing we need to pay attention to is the layout. A Stack might not be the best choice because the title and the close button would overlap if the title gets too long.

Jupi007:
Using a stack was the only way I found to have a centered title, relatively to the dialog width.
Use a row here would look weird because the title widget would be shifted a bit.
I think remember I added a padding to avoid  any overlap.

Feichtmeier:
> Perhaps we could expose `YaruCloseButton` as a separate standalone building block just like we have `YaruBackButton`? :)
> 
> I like the way `AppBar` works. It offers an easy way to lay out the title with leading and trailing actions, and it does the right thing most of the time i.e. it automatically creates the back button for me if appropriate. Yet, it allows me to replace it with something else when more flexibility is needed.
> 
> I'd love `YaruDialogTitle` to be more like `AppBar`. It can create the close button by default but perhaps it's enough to allow replacing it when more customization is needed? Another thing we need to pay attention to is the layout. A Stack might not be the best choice because the title and the close button would overlap if the title gets too long.

sure, so like this plus a
final Widget closeButton
then
if(closeButton == null)
YaruCloseButton()

or is there a close button class?

jpnurmi:
> or is there a close button class?

:+1: that would make it easy to build any kind of title one might want. we started by offering a dialog with a close button. we went down to offering a title with a close button. and now i'm here asking for the close button. what's next? :D

lib/src/controls/yaru_close_button.dart
```diff
@@ -0,0 +1,27 @@
+import 'package:flutter/material.dart';
+import 'package:yaru_icons/yaru_icons.dart';
+
+import 'yaru_icon_button.dart';
+
+class YaruCloseButton extends StatelessWidget {
+  const YaruCloseButton({
+    Key? key,
+    required this.isCloseable,
```
jpnurmi, line 9:
```suggestion
    this.enabled = true,
```

lib/src/controls/yaru_close_button.dart
```diff
@@ -0,0 +1,27 @@
...
+    this.onPressed,
+  }) : super(key: key);
+
+  final bool isCloseable;
```
jpnurmi, line 13:
```suggestion
  final bool enabled;
```

lib/src/controls/yaru_close_button.dart
```diff
@@ -0,0 +1,27 @@
...
+  final Function()? onPressed;
+
+  @override
+  Widget build(BuildContext context) {
+    return YaruIconButton(
+      style: IconButton.styleFrom(
+        fixedSize: const Size.square(34),
+      ),
+      onPressed:
+          isCloseable ? onPressed ?? () => Navigator.of(context).pop() : null,
```
jpnurmi, line 23:
```suggestion
          enabled ? onPressed ?? Navigator.of(context).maybePop : null,
```

Feichtmeier:
> > or is there a close button class?
> 
> +1 that would make it easy to build any kind of title one might want. we started by offering a dialog with a close button. we went down to offering a title with a close button. and now i'm here asking for the close button. what's next? :D

like this?

PR #296, opened 2022-10-13
YaruNavigationRailItem: add tooltip

Jupi007:
I let the tooltip message required, because in the configuration without label, it is the only way to understand what the item mean.

Capture d’écran du 2022-10-13 12-27-16
Capture d’écran du 2022-10-13 12-27-30

Related to previous PR: #243

### DIFF
--- a/lib/src/layouts/yaru_navigation_rail_item.dart
+++ b/lib/src/layouts/yaru_navigation_rail_item.dart
@@ -14,6 +14,7 @@ enum YaruNavigationRailStyle {
 
 const _kSizeAnimationDuration = Duration(milliseconds: 200);
 const _kSelectedIconAnimationDuration = Duration(milliseconds: 250);
+const _kTooltipWaitDuration = Duration(milliseconds: 500);
 
 class YaruNavigationRailItem extends StatefulWidget {
   const YaruNavigationRailItem({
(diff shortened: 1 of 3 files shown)


lib/src/layouts/yaru_navigation_rail_item.dart
```diff
@@ -14,20 +14,23 @@ enum YaruNavigationRailStyle {
...
 const _kSelectedIconAnimationDuration = Duration(milliseconds: 250);
+const _kTooltipWaitDuration = Duration(milliseconds: 500);
 
 class YaruNavigationRailItem extends StatefulWidget {
   const YaruNavigationRailItem({
     super.key,
     this.selected,
     required this.icon,
     required this.label,
+    required this.tooltipMessage,
```
jpnurmi, line 25: Can we call it `tooltip` (see IconButton, FloatingActionButton, PopupMenuButton, ...) and please not make it required? Even though it makes sense to encourage users to set tooltips, we should allow them to build their own tooltips if they prefer. The reason could be as simple as that they want to use rich text, for example.

Jupi007: Okay, you're right :)

jpnurmi (approved):
LGTM :+1:

Issue #287, opened 2022-10-12
Compact and MD layout master scrollareas need to work with touchpads


Feichtmeier:
Since flutter reworked how scrolling works it is now very easy to mess up touchpad scrolling
I also create this issue everytime I use a single child scrollview in software :S 

So @jpnurmi  and @Jupi007 we need to fix the two layout scrolling areas to be either the primary scrollareas (?) or with some other trick :P 
Kinda important bug :warning: 

This only happens if you use the binary outside of vscode (its a wayland / X thing)

jpnurmi:
Is this something that Flutter 3.3 broke? How did you work it around previously?

jpnurmi:
To reproduce from within VS Code:
```
"env": {
  "GDK_BACKEND": "wayland"
}
```

jpnurmi:
flutter/widges/scroll_configuration.dart:
```dart
/// Device types that scrollables should accept drag gestures from by default.
const Set<PointerDeviceKind> _kTouchLikeDeviceTypes = <PointerDeviceKind>{
  PointerDeviceKind.touch,
  PointerDeviceKind.stylus,
  PointerDeviceKind.invertedStylus,
  PointerDeviceKind.trackpad,
  // The VoiceAccess sends pointer events with unknown type when scrolling
  // scrollables.
  PointerDeviceKind.unknown,
};

class ScrollBehavior {
  /// The device kinds that the scrollable will accept drag gestures from.
  ///
  /// By default only [PointerDeviceKind.touch], [PointerDeviceKind.stylus], and
  /// [PointerDeviceKind.invertedStylus] are configured to create drag gestures.
  /// Enabling this for [PointerDeviceKind.mouse] will make it difficult or
  /// impossible to select text in scrollable containers and is not recommended.
  Set<PointerDeviceKind> get dragDevices => _kTouchLikeDeviceTypes;
}
```

yaru_widgets/example/main.dart:
```dart
class TouchMouseStylusScrollBehavior extends MaterialScrollBehavior {
  @override
  Set<PointerDeviceKind> get dragDevices => {
        PointerDeviceKind.touch,
        PointerDeviceKind.mouse,
        PointerDeviceKind.stylus
      };
}
```
We seem to be missing the new `PointerDeviceKind.trackpad` value in the example.

jpnurmi:
> I also create this issue everytime I use a single child scrollview in software :S

hmm, perhaps not the same issue because there's no scroll behavior override in software?

PR #302, opened 2022-10-14
Example: fix touchpad/gesture scrolling on Wayland

jpnurmi:
https://github.com/flutter/flutter/pull/89944 introduced a new `PointerDeviceKind.trackpad` value.

Fixes: #287

### DIFF
--- a/example/lib/main.dart
+++ b/example/lib/main.dart
@@ -118,9 +118,10 @@ class _CompactPage extends StatelessWidget {
 
 class TouchMouseStylusScrollBehavior extends MaterialScrollBehavior {
   @override
-  Set<PointerDeviceKind> get dragDevices => {
-        PointerDeviceKind.touch,
-        PointerDeviceKind.mouse,
-        PointerDeviceKind.stylus
-      };
+  Set<PointerDeviceKind> get dragDevices {
+    return {
+      ...super.dragDevices,
+      PointerDeviceKind.mouse,
+    };
+  }
 }
--- a/example/lib/main.dart
+++ b/example/lib/main.dart
@@ -1,4 +1,3 @@
-import 'package:flutter/gestures.dart';
 import 'package:flutter/material.dart';
 import 'package:provider/provider.dart';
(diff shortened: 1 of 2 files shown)


jpnurmi:
Wait, is the scroll configuration something that should be set from within `YaruCarousel` so that it works as desired in all apps out of the box?

Jupi007:
Hmm, what do you mean exactly?

jpnurmi:
By default, it's not possible to scroll list views and grid views etc. by press-and-drag with the mouse. The example had explicitly enabled that behavior by telling Flutter that mouse is a "drag scrolling" device too. I can see that this was done specifically for the carousel example, so I was wondering if we should do that in the carousel widget then. However, @Feichtmeier is proposing that we get rid of the whole mouse drag override.

Jupi007:
Oh, okay.
Honestly, I think we should totally get rid of the mouse dragging in the carousel. This kind of behaviour is weird on the desktop. 
We already have many possibility of navigation in the carousel with the dots and the navigation buttons.

jpnurmi:
Great, fully agreed! :)

Feichtmeier:
> Oh, okay.
> Honestly, I think we should totally get rid of the mouse dragging in the carousel. This kind of behaviour is weird on the desktop. 
> We already have many possibility of navigation in the carousel with the dots and the navigation buttons.

But I like almost always use my trackpad for swiping pages in the carousel 🎠
You don't meant swiping right? Or do you? :(

Feichtmeier:
https://user-images.githubusercontent.com/15329494/195810254-ac97532a-f6c1-41c3-a2a8-6c11881b7620.mp4

Feichtmeier (approved):
Yes sorry I've forgotten this artefact!

jpnurmi:
Swiping with a touchpad is fine and still works. Press and drag with a mouse no longer scrolls (as expected on desktop).

Jupi007:
Yes, touchpad is fine :)
It is just the mouse dragging which is not something desired on the desktop.

PR #311, opened 2022-10-21
YaruPopupMenuButton: use builder and improve

Feichtmeier:
## Pull request checklist

- [x] I added a before/after/light/dark table if the changes I made could change the components look

| |Before|After|
|-|-|-|
|Light|image|image|
|Dark|image |image|

@jpnurmi as requested the decoration is private and the builder is now the parameter instead of the item list

I also tried to get closer to the outlined button.

### DIFF
--- a/lib/src/controls/yaru_popup_menu_button.dart
+++ b/lib/src/controls/yaru_popup_menu_button.dart
@@ -15,6 +15,7 @@ class YaruPopupMenuButton<T> extends StatelessWidget {
     this.tooltip,
     this.position = PopupMenuPosition.under,
     this.padding = const EdgeInsets.symmetric(horizontal: 5),
+    this.childPadding = const EdgeInsets.symmetric(horizontal: 5),
     this.enabled = true,
   });
 
@@ -26,6 +27,7 @@ class YaruPopupMenuButton<T> extends StatelessWidget {
   final PopupMenuPosition position;
   final List<PopupMenuEntry<T>> Function(BuildContext) itemBuilder;
   final EdgeInsets padding;
+  final EdgeInsets childPadding;
   final bool enabled;
 
   @override
@@ -46,6 +48,7 @@ class YaruPopupMenuButton<T> extends StatelessWidget {
           child: _YaruPopupDecoration(
             child: child,
             padding: padding,
+            childPadding: childPadding,
           ),
         ),
       ),
@@ -59,10 +62,12 @@ class _YaruPopupDecoration extends StatelessWidget {
     super.key,
     required this.child,
     required this.padding,
+    required this.childPadding,
   });
 
   final Widget child;
   final EdgeInsets padding;
+  final EdgeInsets childPadding;
 
   @override
   Widget build(BuildContext context) {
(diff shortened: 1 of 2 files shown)


lib/src/controls/yaru_popup_menu_button.dart
```diff
@@ -9,86 +9,87 @@ class YaruPopupMenuButton<T> extends StatelessWidget {
...
-              child: Icon(
-                YaruIcons.pan_down,
-                size: 20,
+      child: Container(
+        decoration: BoxDecoration(
+          borderRadius: BorderRadius.circular(kYaruButtonRadius),
+          border: Border.all(color: Theme.of(context).dividerColor),
+        ),
+        child: Padding(
+          padding: childPadding,
```
jpnurmi, line 76: Container has a padding property so this can be moved above

lib/src/controls/yaru_popup_menu_button.dart
```diff
@@ -9,86 +9,87 @@ class YaruPopupMenuButton<T> extends StatelessWidget {
...
+          child: Row(
+            mainAxisSize: MainAxisSize.min,
+            children: [
+              Padding(
+                padding: childPadding,
```
jpnurmi, line 81: the same padding is inserted again so you actually get double what you specify?

lib/src/controls/yaru_popup_menu_button.dart
```diff
@@ -9,86 +9,87 @@ class YaruPopupMenuButton<T> extends StatelessWidget {
     super.key,
     required this.initialValue,
     required this.child,
-    required this.items,
+    required this.itemBuilder,
     this.onSelected,
     this.onCanceled,
     this.tooltip,
     this.position = PopupMenuPosition.under,
```
jpnurmi, line 16: the position is not passed to PopupMenuButton

lib/src/controls/yaru_popup_menu_button.dart
```diff
@@ -9,86 +9,87 @@ class YaruPopupMenuButton<T> extends StatelessWidget {
...
+    this.childPadding = const EdgeInsets.symmetric(horizontal: 5),
```
jpnurmi, line 17: by definition, ["padding"](https://www.w3schools.com/css/css_boxmodel.asp) is inside the border - why is this called "child" padding? also, this API is a bit unpredictable because the padding is in fact inserted twice.

lib/src/controls/yaru_popup_menu_button.dart
```diff
@@ -9,86 +9,87 @@ class YaruPopupMenuButton<T> extends StatelessWidget {
...
   });
 
   final T initialValue;
   final Widget child;
-  final List<PopupMenuItem<T>> items;
   final Function(T)? onSelected;
```
jpnurmi, line 22: was `enabled` left out? does it mean there's no way to disable the button?

lib/src/controls/yaru_popup_menu_button.dart
Feichtmeier: No

jpnurmi: image

jpnurmi: image

jpnurmi: perhaps the second padding should be just a sized box to get some spacing between the child and the dropdown icon?

PR #319, opened 2022-10-23
YaruPopupMenuButton: add multi check item

Feichtmeier:
https://user-images.githubusercontent.com/15329494/197391005-98bbfd15-8aa1-440c-b0ae-810c28213279.mp4

### DIFF
--- a/example/lib/pages/popup_page.dart
+++ b/example/lib/pages/popup_page.dart
@@ -1,4 +1,5 @@
 import 'package:flutter/material.dart';
+import 'package:flutter_markdown/flutter_markdown.dart';
 import 'package:yaru_widgets/yaru_widgets.dart';
 
 class PopupPage extends StatefulWidget {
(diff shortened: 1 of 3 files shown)


example/lib/pages/popup_page.dart
```diff
@@ -37,32 +39,48 @@ class _PopupPageState extends State<PopupPage> {
...
-                  ];
-                },
-              );
+          YaruPopupMenuButton<MyEnum>(
+            onSelected: (value) {
+              if (enumSet.contains(value)) {
+                enumSet.remove(value);
+              } else {
+                enumSet.add(value);
+              }
```
jpnurmi, line 69: unfortunately, `onSelected` is meaningless in multi-selection mode. should this code be in `onChanged` to keep `enumSet` in sync and?

```
onChanged: (checked) {
  if (checked) {
    enumSet.add(value);
  } else {
    enumSet.remove(value);
  }
},
```

Feichtmeier: woops yes that slipped through

Just thought that it would be cool to insert code snippets to the yaru widgets example :P Do you know how to force dart syntax in flutter_markdown? ```dart does not work

jpnurmi: Yes, that would be awesome! Hmm, check the Flutter gallery - I recall it has some example snippets

Feichtmeier: ok I take this out of here for now. would be cool to find a generic approach for all pages then

PR #335, opened 2022-10-24
Introduce YaruSwitchButton

jpnurmi:
Similar to YaruCheckButton & YaruRadioButton but with a leading Switch.

| Dark | Light |
|---|---|
| image | image |


jpnurmi:
Motivation: we want to replace the radio buttons in https://github.com/canonical/ubuntu-desktop-installer/pull/1198 with a switch.

PR #337, opened 2022-10-24
YaruCheckedPopupMenuItem: add golden test

jpnurmi:


lib/src/controls/yaru_popup_menu_button.dart
```diff
@@ -159,39 +161,44 @@ class _YaruCheckMark extends StatelessWidget {
...
-            duration: const Duration(
-              milliseconds: duration,
-            ),
-            decoration: BoxDecoration(
-              border: Border.all(
-                color: borderColor,
-                width: 2,
+    return SizedBox(
+      width: _kYaruCheckMarkSize,
+      height: double.infinity,
```
jpnurmi, line 167: had to expand this to full height to be able to center within the expanded height. is there a better way?

Feichtmeier:
> had to expand this to full height to be able to center within the expanded height. is there a better way?

It's fine!

Checked out locally looks good :+1: 
Thank you!

PR #345, opened 2022-10-25
YaruCarousel: allow creating without controller

jpnurmi:
For simpler use cases where the user doesn't need to control the carousel, it's more convenient if `YaruCarousel` internally creates its own controller on demand. It's important to keep in mind that `YaruCarousel` does not own `widget.controller` - only the internal controller that was created if the user did not supply any (i.e. `widget.controller` is null).

## Pull request checklist

- [x] Either this PR does not introduce visual changes, or
- [ ] I run `flutter test --update-goldens` and committed the changes if there were any, or
- [ ] I added before/after/light/dark screenshots if the visual changes I made were not covered by golden tests.

### DIFF
--- a/lib/src/utilities/yaru_carousel.dart
+++ b/lib/src/utilities/yaru_carousel.dart
@@ -7,7 +7,7 @@ class YaruCarousel extends StatefulWidget {
     super.key,
     this.height = 500,
     this.width = 500,
-    required this.controller,
+    this.controller,
     required this.children,
     this.placeIndicator = true,
     this.placeIndicatorMarginTop = 12.0,
@@ -22,7 +22,7 @@ class YaruCarousel extends StatefulWidget {
   /// The width of the children, defaults to 500.0.
   final double width;
 
-  final YaruCarouselController controller;
+  final YaruCarouselController? controller;
 
   /// The list of child widgets shown in the carousel.
   final List<Widget> children;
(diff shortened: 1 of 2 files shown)


Jupi007:
Good idea :+1:

lib/src/utilities/yaru_carousel.dart
```diff
@@ -7,7 +7,7 @@ class YaruCarousel extends StatefulWidget {
     super.key,
     this.height = 500,
     this.width = 500,
-    required this.controller,
+    this.controller,
```
jpnurmi, line 10: i wanted to add `assert(controller == null || controller.pageLength == children.length)` but we'd have to make the constructor non-const... :(

Jupi007: Another solution: remove `controller.pageLength` from the controller constructor, and fill it directly in `initState()` of `_YaruCarouselState`.
Like this we simplify the carousel instantiation, and we ensure that `controller.pageLength == children.length` while keeping a const constructor.

jpnurmi: Thanks, great idea! There seems to be a way to look up YaruCarousel from the scroll notification context (because YaruCarousel is a parent of the scrollable PageView) so we can get the correct value straight from the list of children. :)

PR #348, opened 2022-10-25
Example: update nav icons

Jupi007:
- update to yaru_icons 0.2.6
- use new switch icon
- use filled icons when available (checkbox and radio)

Capture d’écran du 2022-10-25 23-03-46
Capture d’écran du 2022-10-25 23-03-40


jpnurmi (approved):
Nice, LGTM :+1:

Issue #202, opened 2022-09-15
Add yaru switch/check/radio controls


Jupi007:
## Add Yaru controls

Todo list:

- [x] #354 
- [x] #352 
- [x] #353 
- [x] #374
- [x]  #377
- [x] https://github.com/ubuntu/yaru_icons.dart/issues/79
- [x]  #389

jpnurmi:
@Jupi007 @Feichtmeier Sorry to hijack this ticket with something semi-related but do we might want to have the desktop installer's "desktop-like" check and radio buttons in Yaru widgets? They are currently named `CheckButton` and `RadioButton` inspired by GTK. They are very much like `CheckboxListTile` and `RadioListTile` except that they don't have the background highlight of `ListTile`.

[udi-check-radio-buttons.webm](https://user-images.githubusercontent.com/140617/195345140-a2900c0b-ac94-4d2e-8bcd-3c451395ec25.webm)

Jupi007:
Well I want to say why not, it sounds like something generic.
But how does it look in full material?

jpnurmi:
They have nothing theme-specific, they are using the stock Material `Checkbox` and `Radio` indicators:

![image](https://user-images.githubusercontent.com/140617/195348295-2af0959b-7aca-4edd-9468-5b5ae97fed61.png)

Feichtmeier:
> Well I want to say why not, it sounds like something generic. But how does it look in full material?

yes why not

Feichtmeier:
Personally I'm fine with checks and radios
Or at least I think they look better than their Gtk counter part
Maybe, if you decide to work on those, we could somehow merge the looks?

Feichtmeier:
For checks, I realized I created a check for the popup button
I like the even material check mark over the uneven yaru check mark, but the border is weird. How about this?

![grafik](https://user-images.githubusercontent.com/15329494/197694888-f9a6662a-25af-4c86-8c80-acb630c8cdaf.png)
![grafik](https://user-images.githubusercontent.com/15329494/197695070-d67f3af0-707b-4d52-8b4d-41cb9ba48925.png)

Feichtmeier:
those are gtk4 checks
![grafik](https://user-images.githubusercontent.com/15329494/197697695-c6331c12-e9fb-4b15-b393-e1256984fe2c.png)
:shrug:

Feichtmeier:
Hmhmhm @jpnurmi @Jupi007 

[...]

PR #351, opened 2022-10-28
Add YaruCheckbox

### LINKED ISSUE #352: Add YaruCheckbox

Jupi007:
WIP, I will add screenshot later.

- [x] Add basic design
- [x] Add animations
- [x] Add/update golden test
- [x] Add functional test

Related to #202
Fixes #352


Feichtmeier:
The CI fails because some dbus-XXX.deb artifact can't be installed in this ubuntu-latest CI container.
I "fixed" this by downgrading the containers to 22.04
https://github.com/ubuntu-flutter-community/software/pull/446/commits/e1798162e02ce5175262616c64546f846f9f8ee2
For flutter I can not see any negative side effects of this downgrading? @jpnurmi any veto?

jpnurmi:
Moving the CI to 22.04 sounds fine. :+1: `ubuntu-latest` is still pointing to 20.04 which is apparently broken.
- https://github.com/actions/runner-images/issues/6488
- https://github.com/actions/runner-images/issues/6399

Jupi007:
Here is a screencast:

https://user-images.githubusercontent.com/36476595/198849641-2d818862-6886-445c-bec6-a23e901f6091.mp4

Jupi007:
Weird, the golden test fail here, but `flutter test --update-goldens` isn't creating any change locally 🤔

Feichtmeier:
Wow that looks like a huge amount of work! Thank you

I think this looks great and the animations are also nice

The only part I am not convinced about is the form of the checkmark
Can't we leave it evenly formed?
I've noticed that most UI frameworks somehow aligned their checkmarks. Firefox, material and even vanilla framework where the un-even check mark in yaru gnome comes from, changed it back to the even stroked checkmark
grafik

Jupi007:
> Wow that looks like a huge amount of work! Thank you
>
> I think this looks great and the animations are also nice

Thanks a lot 😃
I brained a long time how to implement all this.
I hesitated between reuse some parts of material tooglable class stuff, and redo everything in another way.
And I finally choose the last solution, because like this we will be totally decoupled from their design, and so any material change won't affect us! This also mean, no dirty patch, etc.

> The only part I am not convinced about is the form of the checkmark
> Can't we leave it evenly formed?

Capture d’écran du 2022-10-29 23-23-13

Like this? 🙃
Honestly, after thinking about it, I think you're maybe right 🥲
But I did an adjusted checkmark, it is very sight, but it is not the same as the material one!

madsrh:
This looks great @Jupi007 and very slick animations 😎

Jupi007:
Thanks @madsrh 😃

Feichtmeier:
Hm, merging main should actually also pull in the changes to the CI to use ubuntu-22-04 instead ubuntu-latest which fixes the issue with this dbus.deb 
Wonder why this doesn't happen

Jupi007:
Hmm, both main and this branch are still using ubuntu-latest :)

Feichtmeier:
Check out locally animations are really :+1: nice

Just one question, is the size change intended?
grafik

I am not against or pro just wanted to be sure this does not have any negative side effects when we exchange material checkboxes with yarucheckboxes in "all" our apps : D

Jupi007:
There should be no problem :)

https://user-images.githubusercontent.com/36476595/198971099-89244ac8-dec4-49c0-90d2-536be8014491.mp4

The material checkbox got a transparent activable padding, the yaru one also have one, but smaller 😉

PR #358, opened 2022-11-03
Master detail controller

d-loose:
Adds a simple controller to `YaruMasterDetailPage` that allows to select the currently selected index.

## Pull request checklist

- [x] Either this PR does not introduce visual changes, or
- [ ] I run `flutter test --update-goldens` and committed the changes if there were any, or
- [ ] I added before/after/light/dark screenshots if the visual changes I made were not covered by golden tests.

### DIFF
--- a/lib/src/layouts/yaru_landscape_layout.dart
+++ b/lib/src/layouts/yaru_landscape_layout.dart
@@ -19,6 +19,7 @@ class YaruLandscapeLayout extends StatefulWidget {
     required this.pageMinWidth,
     this.onLeftPaneWidthChange,
     this.appBar,
+    this.controller,
   });
 
   /// The total number of pages.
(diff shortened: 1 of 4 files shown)


Feichtmeier:
Hi @d-loose 

Does this also allow the outer navigation to be controlled from inner pages?
We got a feature request on the software store that is related to this eventually
https://github.com/ubuntu-flutter-community/software/issues/254

d-loose:
No, with the controller you can only change the selected index.

@jpnurmi mentioned this issue to me. It's already possible to do that by changing the state in `onSelected` and making the detail page rebuild accordingly (it seems like the state of the `ExplorePage` in Ubuntu Software is contained within its view model, so it might not be that simple in this case, unless you move it higher up).

Feichtmeier:
@d-loose returning from the :ramen: talk :laughing: ... I can not see a way to apply this to the compact layout :thinking: 
Could eventually add this also to compact layout?

d-loose:
Ah, I misunderstood you earlier, I thought you were talking about the portrait layout. I'll look into it! There are still some issues with this very simplistic approach anyways:
- the controller doesn't change its value when the user manually selects another item
- using the controller to navigate to the same item twice (with some user interaction in between) doesn't work, because the controller still holds the same value and doesn't notify its listeners if it's set to the same value again

To make everything work smoothly, every action should involve the controller - just like flutter's `Tab` widgets

Feichtmeier:
Okay :) Ty!

PR #367, opened 2022-11-09
Added snap packaging for yaru-widgets-example

kenvandine:
No visual changes, adds snap packaging for "yaru-widgets-example" which we can automatically publish on build to keep an easy way for developers to visualize the widgets.


Jupi007:
Thanks!
This is a really good idea.

kenvandine:
Any preferences for the snap name?  I started with yaru-widgets-example, thoughts on that?

Feichtmeier:
@kenvandine thanks! nice idea
yeah that sounds about right!

PR #379, opened 2022-11-11
Fix YaruCarouselController.startTimer()

jpnurmi:
No visual changes.

Don't schedule a timer when no longer attached. This fixes the following exception that occurred while randomly browsing the example.

```
[ERROR:flutter/runtime/dart_vm_initializer.cc(41)] Unhandled Exception: 'package:flutter/src/widgets/scroll_controller.dart':
Failed assertion: line 107 pos 12: '_positions.isNotEmpty': ScrollController not attached to any scroll views.
```


Jupi007 (approved):
Thanks :D

PR #381, opened 2022-11-11
YaruSection: simplify headline & remove headerWidget

jpnurmi:
It's easy enough to wrap stuff into a Row when needed, and then you'll have full control over the main & cross-axis alignment etc. ;-)

## Pull request checklist

- [x] this PR does not introduce visual changes

### DIFF
--- a/lib/src/pages/yaru_section.dart
+++ b/lib/src/pages/yaru_section.dart
@@ -6,17 +6,17 @@ class YaruSection extends StatelessWidget {
   /// [Widgets] as children.
   const YaruSection({
     super.key,
-    this.title,
+    this.headline,
     required this.child,
     this.width,
     this.height,
     this.padding = const EdgeInsets.all(8.0),
-    this.titlePadding = const EdgeInsets.all(8.0),
+    this.headlinePadding = const EdgeInsets.all(8.0),
     this.margin,
   });
 
   /// Widget that is placed above the `child`.
-  final Widget? title;
+  final Widget? headline;
 
   /// The child widget inside the section.
   final Widget child;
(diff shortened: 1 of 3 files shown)


lib/src/pages/yaru_section.dart
```diff
@@ -27,16 +27,13 @@ class YaruSection extends StatelessWidget {
...
   final double? height;
 
-  /// Aligns the widget horizontally along with headline.
-  ///
-  /// Both `headline` and `headerWidget` will be aligned horizontally
-  /// with [mainAxisAlignment] as [MainAxisAlignment.spaceBetween].
-  final Widget? headerWidget;
-
   /// The padding between the section border and its [child] which defaults to
-  /// `EdgeInsets.only(all: 8.0)`.
```
jpnurmi, line 37: only all... oops :man_facepalming:

Feichtmeier:
Since you kept the default text style it should be okay I think
Because this was the main pattern

--------------
| Headline |

| Children  |
----------------

Feichtmeier:
Since you kept the default text style it should be okay I think
Because this was the main pattern

--------------
| Headline |

| Children  |
----------------

Edit: could you check if there is still the same header style?

jpnurmi:
Yes, I kept the same default text style. There are no visual changes in the example:
image

jpnurmi:
Or would you prefer to keep the original `headline` property name?

Feichtmeier:
Ok I don't have a strong opinion 🐱 go with whatever you prefer then

jpnurmi:
My English is not good enough to understand the real difference between "title" and "headline". I'm just used to one of them in Flutter... :rofl:

PR #399, opened 2022-11-19
Fix wrong label wrap in page layouts

### LINKED ISSUE #259: YaruCompactLayout: wrong label wrap

Jupi007:
- Create `YaruOneLineText` widget.
- Add missing comments in `YaruNavigationRailItem`.

Fixes #259

## Pull request checklist

- [x]  I added before/after/light/dark screenshots if the visual changes I made were not covered by golden tests.
    
| |Before|After|
|-|-|-|
|Light| Capture d’écran du 2022-11-19 11-14-38 | Capture d’écran du 2022-11-19 11-14-32 |
|Dark| Capture d’écran du 2022-11-19 11-27-23 | Capture d’écran du 2022-11-19 11-27-11 |

**Edit:** sorry for the push mess, I force pushed by mistake an empty commit 🤕

### DIFF
--- a/example/lib/example.dart
+++ b/example/lib/example.dart
@@ -35,7 +35,7 @@ class _ExampleState extends State<Example> {
   Widget build(BuildContext context) {
     final model = context.watch<ExampleModel>();
     final configItem = PageItem(
-      titleBuilder: (context) => const Text('Layout'),
+      titleBuilder: (context) => YaruOneLineText('Layout'),
       tooltipMessage: 'Layout',
       snippetUrl:
           'https://raw.githubusercontent.com/ubuntu/yaru_widgets.dart/main/lib/src/layouts/yaru_landscape_layout.dart',
--- a/example/lib/example_page_items.dart
+++ b/example/lib/example_page_items.dart
@@ -1,5 +1,6 @@
 import 'package:flutter/material.dart';
 import 'package:yaru_icons/yaru_icons.dart';
+import 'package:yaru_widgets/yaru_widgets.dart';
 import 'pages/banner_page.dart';
 import 'pages/carousel_page.dart';
 import 'pages/check_button_page.dart';
@@ -38,7 +39,7 @@ class PageItem {
 
 final examplePageItems = <PageItem>[
   PageItem(
-    titleBuilder: (context) => const Text('Controls'),
+    titleBuilder: (context) => YaruOneLineText('Controls'),
     tooltipMessage: 'Controls',
     snippetUrl:
         'https://raw.githubusercontent.com/ubuntu/yaru_widgets.dart/main/example/lib/pages/controls_page.dart',
@@ -46,7 +47,7 @@ final examplePageItems = <PageItem>[
     iconBuilder: (context, selected) => const Icon(YaruIcons.games),
   ),
   PageItem(
-    titleBuilder: (context) => const Text('YaruBanner'),
+    titleBuilder: (context) => YaruOneLineText('YaruBanner'),
     tooltipMessage: 'YaruBanner',
     snippetUrl:
         'https://raw.githubusercontent.com/ubuntu/yaru_widgets.dart/main/example/lib/pages/banner_page.dart',
@@ -56,7 +57,7 @@ final examplePageItems = <PageItem>[
         : const Icon(YaruIcons.image),
   ),
   PageItem(
-    titleBuilder: (context) => const Text('YaruCarousel'),
+    titleBuilder: (context) => YaruOneLineText('YaruCarousel'),
     tooltipMessage: 'YaruCarousel',
     snippetUrl:
         'https://raw.githubusercontent.com/ubuntu/yaru_widgets.dart/main/example/lib/pages/carousel_page.dart',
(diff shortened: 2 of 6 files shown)


Feichtmeier:
I'm not sure about this one
It's a good fix no question
But it's also adding complexity to such a simple thing as a text.
Couldn't we alternatively as this to the software app if needed? Or could we create an extension? No, this would affect all text. Hm let @jpnurmi have an opinion on this one please

Jupi007:
I completely agree with you.
I really don't like this solution, but it is the only one which works 😐
And as each app which use the nav rail will have this problem, I think we should add a solution here and not in each app.

I would really like to know what you think of this @jpnurmi

Feichtmeier:
What about a sizedbox ? Shouldn't a sizedbox or constrained box be enough to tell the text when to break?

Jupi007:
No, I really tried a lot of stuff. The only way to do an ellipsis in a word, is to drop all the space in the string by unbreakable characters 😕

jpnurmi:
the string replacement operation could be a one-liner String extension but i don't think it belongs to yaru and i'm still unsure if this really applies to all languages because then this would be the default behavior in flutter... :)

jpnurmi:
Actually, is there any Flutter issue about this? Maybe it's worth reporting?

Jupi007:
I hadn't thought of an issue with Flutter, but it is indeed a problem, and we should report it 👍
(and not do dirty fix 😅)

I will report this today, if this isn't already an existent issue.

jpnurmi:
I don't understand the truncation logic in Flutter...

[Screencast from 2022-11-25 11-06-03.webm](https://user-images.githubusercontent.com/140617/203957612-6a2e4a7f-a389-43cd-831a-b658887ed30e.webm)

```dart

void main() {
  runApp(
    const MaterialApp(
      home: Scaffold(
        body: Padding(
          padding: EdgeInsets.all(48.0),
          child: DecoratedBox(
            decoration: BoxDecoration(color: Colors.grey),
            child: Text(
              'Lorem ipsum dolor sit amet',
              overflow: TextOverflow.ellipsis,
              style: TextStyle(fontSize: 64),
            ),
          ),
        ),
      ),
    ),
  );
}
```

Jupi007:
Wait, why does it works here.
And it works... but not for the last word of the sentence 😅

jpnurmi:
There are some strange jumps like the first three letters of "dolor..." => "dolo..." => "dol..." => "..."

Qt behaves a little bit better but also has a strange jump for the last word:

[Screencast from 2022-11-25 11-23-04.webm](https://user-images.githubusercontent.com/140617/203962270-21b56f17-d1ce-4f7c-be3b-60e4f713b3c6.webm)

```qml
import QtQuick 2.15
import QtQuick.Window 2.15

Window {
    width: 640
    height: 480
    visible: true

    Text {
        text: "Lorem ipsum dolor sit amet"
        elide: Text.ElideRight
        font.pixelSize: 64
        anchors.top: parent.top
        anchors.left: parent.left
        anchors.right: parent.right
        anchors.margins: 48

        Rectangle {
            z: -1
            anchors.fill: parent
            color: "gray"
        }
    }
}
```

d-loose:
For the last word it makes sense to only show ellipsis after removing at least three characters - otherwise you'd actually make the string longer

jpnurmi:
Haha right, I'm glad I didn't post that somewhere in public! :upside_down_face: 

Last but not least, here's GTK - on par with Qt:

[Screencast from 2022-11-25 11-30-48.webm](https://user-images.githubusercontent.com/140617/203963901-811a51fd-da7b-44e6-9351-53a561c9cc35.webm)

jpnurmi:
I'm starting to see the pattern now...

[Screencast from 2022-11-25 11-39-40.webm](https://user-images.githubusercontent.com/140617/203965711-5d092f3f-94be-4e6d-ae77-d12be5322328.webm)

d-loose:
https://github.com/flutter/flutter/issues/18761

Jupi007:
And someone mentioned the same workaround as this PR: https://github.com/flutter/flutter/issues/18761#issuecomment-1248988493

jpnurmi:
Great! In that case, I'd definitely not create a Yaru _widget_ as a workaround but have a one-liner String extension somewhere in the app. It probably already exists somewhere but if we can't find any ready-made extension package, we could even try contributing it to one of those (dartx, string_extensions, ...)

```dart
extension Nbsp on String {
  String get nbsp => replaceAll(' ', '\u{000A0}');
}
```

Jupi007:
I agree that this shouldn't be a Yaru widget.
But so, use one of our layout will require another package, and I think it is a bit annoying 😕

PR #402, opened 2022-11-19
Improve togglable widget style

### LINKED ISSUE #377: Unchecked checkboxes and radio buttons may look disabled to some
The new checkboxes and radio buttons are filled with gray when unchecked, which somehow resembles how some themes and UI toolkits dim disabled checkboxes and radio buttons.

Personally, I got confused while changing the desktop installer to use the new Yaru checkboxes and radio buttons. I thought I did something wrong that caused the checkboxes and radio buttons to become disabled so I thought I'd raise a ticket for consideration because there's a chance that someone else might also misinterpret them. :) Perhaps a subtle frame would help to make it look more interactive?

| Before | After |
|---|---|
| Screenshot from 2022-11-11 14-16-31 | Screenshot from 2022-11-11 14-14-41 |

Ref: canonical/ubuntu-desktop-installer#1223

Jupi007:
I was unsatisfied of the switch dot look, so I did something different from my mockup.

Fixes #377

## Pull request checklist

- [x]  I ran `flutter test --update-goldens` and committed the changes if there were any

| | Before | After |
|-|-|-|
|Light| Capture d’écran du 2022-11-19 15-18-08 | Capture d’écran du 2022-11-19 15-19-17 |
|Dark| Capture d’écran du 2022-11-19 15-18-24 | Capture d’écran du 2022-11-19 15-19-33 |

CC @jpnurmi @elioqoshi @madsrh @Feichtmeier


Jupi007:
In video:

https://user-images.githubusercontent.com/36476595/202855869-2d8e6df2-3b50-4ca9-9fe7-3739130da7d9.mp4

https://user-images.githubusercontent.com/36476595/202855774-68366181-448c-4e5e-81e9-b9e752ce1980.mp4

madsrh:
@Jupi007 Are you sure you don't want a white dot for active switches in the off state? Seems to me that this gray one easily could be mistaken for a deactivated switch 🤷‍♂️ 

I like the "stroke" / outline, but I wonder if it's a bit too strong
image

Jupi007:
I agree with you @madsrh 
So what do you think of something like this:

| Light | Dark |
| ------ | ------- |
| Capture d’écran du 2022-11-19 17-52-39 | Capture d’écran du 2022-11-19 17-56-22 |

Feichtmeier:
yes, this is almost perfect. maybe now only remove the bg of the unchecked not-disabled dark boxes? I know this sounds weird since I said let's try this mix but sometimes you need to see how it goes ^.^

madsrh:
> So what do you think of something like this:

That is perfect 👌 

> yes, this is almost perfect. maybe now only remove the bg of the unchecked not-disabled dark boxes?

Like this? 🤔 
image

Jupi007:
It is good like this?
I darken a very little the checkradio background (black with 0.05 of opacity) to increase the contrast:

| Before | After |
| ------ | ------- |
| Capture d’écran du 2022-11-19 17-56-22 | image |

I think it looks very good like this 😃

Feichtmeier:
I would make the background "transparent". I fear they stick out too much because all the buttons are outlined

Jupi007:
Okay, well we should also use transparent on the light theme so:

| Light | Dark |
| ------ | ------- |
| Capture d’écran du 2022-11-19 19-01-16 | Capture d’écran du 2022-11-19 19-01-29 |

Jupi007:
@madsrh Have I your go-ahead for this version? 👆

madsrh:
Yep 👍

Feichtmeier (approved):

PR #403, opened 2022-11-19
Add YaruMasterDetailPaneLayoutDelegate

### LINKED ISSUE #393: Add YaruMasterDetailResizeDelegate
What about grouping anything resizing-related as required properties into a "constraints" or "delegate" type of object? It's arguably less convenient for users but forcing them to think about the size constraints prevents such surprises that may happen if we allow them to just set one property and they expect it to magically do the right thing (whatever it means in their case).

```dart
YaruMasterDetailPage(
  resizeDelegate: YaruMasterResizeDelegate( // or "Constraints"?
    minPaneWidth: 123,
    minPageWidth: 456,
  ),
),
```

_Originally posted by @jpnurmi in https://github.com/ubuntu/yaru_widgets.dart/issues/266#issuecomment-1274560891_

Jupi007:
Add `YaruMasterDetailPaneLayoutDelegate`, which controls a `YaruMasterDetailPage` pane with defined parameters.
This force the user to think about layout size constraints of.

Fixes #393

## Pull request checklist

### DIFF
--- a/lib/src/layouts/yaru_master_detail_page.dart
+++ b/lib/src/layouts/yaru_master_detail_page.dart
@@ -7,6 +7,8 @@ import 'yaru_master_detail_theme.dart';
 import 'yaru_master_tile.dart';
 import 'yaru_portrait_layout.dart';
 
+const _kDefaultPaneWidth = 280.0;
+
 typedef YaruMasterDetailBuilder = Widget Function(
   BuildContext context,
   int index,
@@ -50,7 +52,8 @@ class YaruMasterDetailPage extends StatefulWidget {
     required this.length,
     required this.tileBuilder,
     required this.pageBuilder,
-    required this.layoutDelegate,
+    this.layoutDelegate =
+        const YaruMasterFixedPaneDelegate(paneWidth: _kDefaultPaneWidth),
     this.appBar,
     this.initialIndex,
     this.onSelected,


jpnurmi:
This is great! Is there any way to come up with a reasonable fixed default? Was the former `leftPaneWidth` always a required property?

Feichtmeier:
> This is great! Is there any way to come up with a reasonable fixed default? Was the former `leftPaneWidth` always a required property?

Yeah it was :D I also vote for a default :) The less to configure the better

Jupi007:
I added a default pane width to both delegate :)

lib/src/layouts/yaru_master_detail_page.dart
```diff
@@ -49,10 +50,7 @@ class YaruMasterDetailPage extends StatefulWidget {
     required this.length,
     required this.tileBuilder,
     required this.pageBuilder,
-    required this.leftPaneWidth,
-    this.allowLeftPaneResize = true,
-    this.leftPaneMinWidth = 175.0,
-    this.pageMinWidth = kYaruMasterDetailBreakpoint / 2,
+    required this.layoutDelegate,
```
jpnurmi, line 53:
```
this.layoutDelegate = const YaruMasterFixedPaneDelegate(paneWidth: _kDefaultPaneWidth),
```

lib/src/layouts/yaru_master_detail_layout_delegate.dart
```diff
@@ -0,0 +1,85 @@
...
+    return paneWidth;
+  }
+}
+
+/// Controls a [YaruMasterDetailPage] pane with a resizable width.
+class YaruMasterResizablePaneDelegate
+    implements YaruMasterDetailPaneLayoutDelegate {
+  /// Controls a [YaruMasterDetailPage] pane with a resizable width.
+  const YaruMasterResizablePaneDelegate({
+    this.initialPaneWidth = _kDefaultPaneWidth,
```
jpnurmi, line 50: remove the default value

lib/src/layouts/yaru_master_detail_layout_delegate.dart
```diff
@@ -0,0 +1,85 @@
...
+    required double? candidatePaneWidth,
+  });
+}
+
+/// Controls a [YaruMasterDetailPage] pane with a fixed width.
+class YaruMasterFixedPaneDelegate
+    implements YaruMasterDetailPaneLayoutDelegate {
+  /// Controls a [YaruMasterDetailPage] pane with a fixed width.
+  const YaruMasterFixedPaneDelegate({
+    this.paneWidth = _kDefaultPaneWidth,
```
jpnurmi, line 22: remove the default value

jpnurmi (commented):
i'd keep the delegate properties required and assign a default layout delegate instead. if the user is not happy with the default behavior, they can create a custom delegate but then they are forced to enter all relevant values to avoid surprises.

Jupi007:
Ohh, okay!
Good idea :+1:

jpnurmi (approved):
Thanks, LGTM :+1:

Jupi007:
And thank you very much for the idea! 🙂

PR #404, opened 2022-11-20
Improve togglable contrast

Jupi007:
- Increase togglable border contrast ;
- Use `onSurface` as background to avoid hover/focus indicator to be visible by transparency.

CC @madsrh 

Capture d’écran du 2022-11-20 11-51-35
Capture d’écran du 2022-11-20 11-51-26

## Pull request checklist

- [x]  I ran `flutter test --update-goldens` and committed the changes if there were any


Feichtmeier (approved):
Thank you 🙇

PR #414, opened 2022-11-24
Yaru layout index controller

d-loose:
WIP
Adds ~~`YaruLayoutIndexController`~~ `YaruPageController` that controls
- [x] `YaruMasterDetailPage`
    * [x] `YaruLandscapeLayout`
    * [x] `YaruPortraitLayout`
- [x] `YaruCompactLayout` 

## Pull request checklist

    | |Before|After|
    |-|-|-|
    |Light| | |
    |Dark| | |

### DIFF
--- a/lib/src/layouts/yaru_compact_layout.dart
+++ b/lib/src/layouts/yaru_compact_layout.dart
@@ -79,7 +79,10 @@ class _YaruCompactLayoutState extends State<YaruCompactLayout> {
   @override
   void didUpdateWidget(covariant YaruCompactLayout oldWidget) {
     super.didUpdateWidget(oldWidget);
-    if (widget.controller != oldWidget.controller) _updatePageController();
+    if (widget.controller != oldWidget.controller) {
+      oldWidget.controller?.removeListener(_pageControllerCallback);
+      _updatePageController();
+    }
   }
 
   void _updatePageController() {
--- a/lib/src/layouts/yaru_compact_layout.dart
+++ b/lib/src/layouts/yaru_compact_layout.dart
@@ -53,7 +53,6 @@ class YaruCompactLayout extends StatefulWidget {
 }
 
 class _YaruCompactLayoutState extends State<YaruCompactLayout> {
-  late int _index;
   late final int _length;
 
   late final ScrollController _scrollController;
@@ -65,7 +64,6 @@ class _YaruCompactLayoutState extends State<YaruCompactLayout> {
     _length = widget.length ?? widget.controller!.length;
     _scrollController = ScrollController();
     _updatePageController();
-    _index = max(_pageController.initialIndex, widget.initialIndex ?? 0);
   }
 
   @override
@@ -79,6 +77,9 @@ class _YaruCompactLayoutState extends State<YaruCompactLayout> {
   @override
   void didUpdateWidget(covariant YaruCompactLayout oldWidget) {
     super.didUpdateWidget(oldWidget);
+    if (widget.length != oldWidget.length) {
+      _length = widget.length ?? widget.controller!.length;
+    }
     if (widget.controller != oldWidget.controller) {
       oldWidget.controller?.removeListener(_pageControllerCallback);
       _updatePageController();
@@ -86,14 +87,13 @@ class _YaruCompactLayoutState extends State<YaruCompactLayout> {
   }
 
   void _updatePageController() {
-    _pageController = widget.controller ?? YaruPageController(length: _length);
+    _pageController =
+        widget.controller ?? YaruPageController(length: widget.length!);
     _pageController.addListener(_pageControllerCallback);
   }
 
   void _pageControllerCallback() {
-    if (_pageController.index != _index) {
-      setState(() => _index = _pageController.index);
-    }
+    setState(() {});
   }
 
   @override
@@ -134,7 +134,7 @@ class _YaruCompactLayoutState extends State<YaruCompactLayout> {
         child: ConstrainedBox(
           constraints: BoxConstraints(minHeight: constraint.maxHeight),
           child: YaruNavigationRail(
-            selectedIndex: _index,
+            selectedIndex: max(_pageController.index, 0),
             onDestinationSelected: _onTap,
             length: _length,
             itemBuilder: widget.itemBuilder,
@@ -150,6 +150,8 @@ class _YaruCompactLayoutState extends State<YaruCompactLayout> {
 
   Widget _buildPageView(BuildContext context) {
     final theme = YaruCompactLayoutTheme.of(context);
+    final index = max(_pageController.index, 0);
+
     return Expanded(
       child: Theme(
         data: Theme.of(context).copyWith(
@@ -158,9 +160,9 @@ class _YaruCompactLayoutState extends State<YaruCompactLayout> {
         child: Navigator(
           pages: [
             MaterialPage(
-              key: ValueKey(_index),
-              child: _length > _index
-                  ? widget.pageBuilder(context, _index)
+              key: ValueKey(index),
+              child: _length > index
+                  ? widget.pageBuilder(context, index)
                   : widget.pageBuilder(context, 0),
             ),
           ],
--- a/lib/src/layouts/yaru_compact_layout.dart
+++ b/lib/src/layouts/yaru_compact_layout.dart
@@ -21,9 +21,10 @@ class YaruCompactLayout extends StatefulWidget {
     required this.length,
     required this.itemBuilder,
     required this.pageBuilder,
+    this.initialIndex,
     this.onSelected,
     this.controller,
-  });
+  }) : assert(initialIndex == null || controller == null);
 
   /// The total number of pages.
   final int length;
@@ -36,6 +37,9 @@ class YaruCompactLayout extends StatefulWidget {
 
   /// A builder that is called for each page to build its content.
   final IndexedWidgetBuilder pageBuilder;
+
+  /// The index of the initial page to show.
+  final int? initialIndex;
 
   /// Called when the user selects a page.
   final ValueChanged<int>? onSelected;
@@ -57,7 +61,7 @@ class _YaruCompactLayoutState extends State<YaruCompactLayout> {
   void initState() {
     _scrollController = ScrollController();
     _updatePageController();
-    _index = max(_pageController.initialIndex, 0);
+    _index = max(_pageController.initialIndex, widget.initialIndex ?? 0);
     super.initState();
   }
 
--- a/lib/src/layouts/yaru_compact_layout.dart
+++ b/lib/src/layouts/yaru_compact_layout.dart
@@ -53,15 +53,14 @@ class YaruCompactLayout extends StatefulWidget {
 }
 
 class _YaruCompactLayoutState extends State<YaruCompactLayout> {
-  late final int _length;
-
   late final ScrollController _scrollController;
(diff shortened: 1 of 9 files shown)


jpnurmi (commented):
naming idea: `YaruPageController` :slightly_smiling_face:

d-loose:
Naming things is always the hardest part :D

jpnurmi:
I wish `YaruCompactLayout` would be named something like `YaruNavigationPage` to be consistent with `YaruMasterDetailPage` and `YaruNavigationRail` but it would break the software app once again... :thinking:

Feichtmeier:
Breaking software is not the problem
I just thought about a future where we might want to readd the layout builder and the bottom bar again
But then this could also be done inside the app so feel free to rename it ;P

jpnurmi:
Sorry for repeatedly hijacking random PRs. :slightly_smiling_face: Let's continue the discussion here:
- https://github.com/ubuntu/yaru_widgets.dart/issues/416

lib/src/layouts/yaru_compact_layout.dart
```diff
@@ -34,34 +37,56 @@ class YaruCompactLayout extends StatefulWidget {
...
+    _scrollController.dispose();
+    _pageController.removeListener(_pageControllerCallback);
+    if (widget.controller == null) _pageController.dispose();
     super.dispose();
   }
 
+  @override
+  void didUpdateWidget(covariant YaruCompactLayout oldWidget) {
+    super.didUpdateWidget(oldWidget);
+    if (widget.controller != oldWidget.controller) _updatePageController();
```
jpnurmi, line 75: missing `removeListener()` call on the old controller if not null

lib/src/layouts/yaru_master_detail_page.dart
```diff
@@ -55,7 +57,6 @@ class YaruMasterDetailPage extends StatefulWidget {
     this.layoutDelegate =
         const YaruMasterFixedPaneDelegate(paneWidth: _kDefaultPaneWidth),
     this.appBar,
-    this.initialIndex,
```
jpnurmi, line 58: same as `YaruCompactLayout` - we could keep this for convenience?

lib/src/layouts/yaru_compact_layout.dart
```diff
@@ -18,8 +21,8 @@ class YaruCompactLayout extends StatefulWidget {
     required this.length,
```
jpnurmi, line 21: both `YaruCompactLayout` _and_ `YaruPageController` have required `length`

lib/src/layouts/yaru_compact_layout.dart
```diff
@@ -18,8 +21,8 @@ class YaruCompactLayout extends StatefulWidget {
...
     required this.itemBuilder,
     required this.pageBuilder,
-    this.initialIndex = 0,
```
jpnurmi, line 21: this could be convenient to keep for the internally constructed controller. we could assert that only `initialIndex` or `controller` but not both are set. what do you think?

lib/src/layouts/yaru_master_detail_page.dart
```diff
@@ -81,43 +82,39 @@ class YaruMasterDetailPage extends StatefulWidget {
...
+    super.dispose();
   }
 
   @override
   void didUpdateWidget(covariant YaruMasterDetailPage oldWidget) {
     super.didUpdateWidget(oldWidget);
-    if (widget.initialIndex != oldWidget.initialIndex) {
-      _index = widget.initialIndex ?? -1;
-    }
+    if (widget.controller != oldWidget.controller) _updateController();
```
jpnurmi, line 117: missing `removeListener()` call on the old controller if not null

lib/src/layouts/yaru_compact_layout.dart
d-loose: Good point, it's probably more convenient to simply pass an argument than creating your own controller if all you need to do is set the initial index :+1:

lib/src/layouts/yaru_master_detail_page.dart
d-loose: `YaruMasterDetailPage` never adds any listener - that's done by either `YaruLandscapeLayout` or `YaruPortraitLayout`, which remove their listeners in `dispose()`. Is there anything that could go wrong?

lib/src/layouts/yaru_compact_layout.dart
d-loose: Is that a problem? What would be a better solution?
Flutter does this in a similar way with `TabController` and the `Tab*` widgets (except that the widgets get a list of children with a fixed length - here we use builder functions that take an index instead of lists, so we need the length)

lib/src/layouts/yaru_master_detail_page.dart
d-loose: :+1:

lib/src/layouts/yaru_compact_layout.dart
d-loose: That's probably the best option - thanks

lib/src/layouts/yaru_master_detail_page.dart
jpnurmi: oh, i may have mixed up the layout names :)

jpnurmi: actually, are they both missing `didUpdateWidget()` here? can't the controller be different at init and dispose time if someone creates a controller inline and the whole thing gets rebuilt with a different controller?

d-loose: Yes, that can happen. Added `didUpdateWidget` to both - hope it works correctly now

jpnurmi: looks good i think

jpnurmi (commented):
my head hearts with all these layouts, indexes, controllers, and listeners but i think it looks good now :+1:

lib/src/layouts/yaru_compact_layout.dart
```diff
@@ -35,33 +40,62 @@ class YaruCompactLayout extends StatefulWidget {
...
-    _controller.dispose();
+    _scrollController.dispose();
+    _pageController.removeListener(_pageControllerCallback);
+    if (widget.controller == null) _pageController.dispose();
     super.dispose();
   }
 
+  @override
+  void didUpdateWidget(covariant YaruCompactLayout oldWidget) {
+    super.didUpdateWidget(oldWidget);
```
jpnurmi, line 81: the widget has been rebuilt and its `length` property may have changed so we need to re-sync `_length`

lib/src/layouts/yaru_compact_layout.dart
```diff
@@ -35,33 +40,62 @@ class YaruCompactLayout extends StatefulWidget {
...
+    if (widget.controller != oldWidget.controller) {
+      oldWidget.controller?.removeListener(_pageControllerCallback);
+      _updatePageController();
+    }
+  }
+
+  void _updatePageController() {
+    _pageController = widget.controller ?? YaruPageController(length: _length);
```
jpnurmi, line 89: should it pass `widget.length` because `_length` may be the previous controller's length?

lib/src/layouts/yaru_compact_layout.dart
```diff
@@ -35,33 +40,62 @@ class YaruCompactLayout extends StatefulWidget {
...
 
+  /// An optional controller that can be used to navigate to a specific index.
+  final YaruPageController? controller;
+
   @override
   State<YaruCompactLayout> createState() => _YaruCompactLayoutState();
 }
 
 class _YaruCompactLayoutState extends State<YaruCompactLayout> {
   late int _index;
```
jpnurmi, line 56: do we need a separate `_index` anymore because a controller always exists so maybe we could pass `_pageController.index` to the layout? the less we have these state variables the better because they all need to be synced in `didUpdateWidget()` :)

lib/src/layouts/yaru_master_detail_page.dart
```diff
@@ -88,36 +91,40 @@ class YaruMasterDetailPage extends StatefulWidget {
...
+      YaruPageController(
+        length: _length,
+        initialIndex: widget.initialIndex ?? -1,
+      );
 
   @override
   void initState() {
     super.initState();
-    _index = widget.initialIndex ?? -1;
+    _length = widget.length ?? widget.controller!.length;
```
jpnurmi, line 114: re-sync in `didUpdateWidget()` or make it a convenience getter instead

lib/src/layouts/yaru_compact_layout.dart
d-loose: :+1:

lib/src/layouts/yaru_compact_layout.dart
d-loose: :+1:

lib/src/layouts/yaru_compact_layout.dart
d-loose: I kept it to deal with a negative index (which is only really needed in `YaruPortraitLayout`), but it's simpler to deal with this in the appropriate places than taking care of another state variable :+1:

lib/src/layouts/yaru_master_detail_page.dart
d-loose: Using a getter is a good idea - is there any reason this wouldn't work in the compact layout as well?

jpnurmi: i like the getter too and i _thought_ i had proposed it above for CL. i don't know what happened but sorry for the confusion - i kept mixing up the classes while reviewing

jpnurmi (commented):
now that we always have a controller instance, we could let `AnimatedBuilder` take care of rebuilds:

```dart
AnimatedBuilder(
  animation: _pageController,
  builder: (context, child) => YaruFooBarPageOrLayout(
    index: _pageController.index
  ),
),
```

this way, we could get rid of the manual `add/removeListener()` and `setState()` calls.

d-loose:
Do we want to keep `YaruPortraitLayout` and `YaruLandscapeLayout` available as stand-alone layout widgets with their own controllers? Using the `AnimatedBuilder` like this in `YaruMasterDetailPage` would make `YaruPortraitLayout` and `YaruLandscapeLayout` stateless and dependent on a parent widget that handles interactions.

jpnurmi:
We should be free to change `YaruPortraitLayout` and `YaruLandscapeLayout` as we wish. They are not exported in `lib/yaru_widgets.dart` because they are supposed to be internal implementation details. Do you think they would be valuable as public classes?

d-loose:
Maybe there are situations where you wouldn't want the layout to change, but I guess this can be controlled by setting the breakpoint to zero or infinity.
But I like the idea of moving all control to the MD page, it simplifies things a lot.

jpnurmi:
Is this good to go as it is now? We could leave using `AnimatedBuilder` and the consequent changes as a separate step.

d-loose:
Yes, should be good to go - let's do it in a separate PR. There are still some things to consider with the `AnimatedBuilder` (navigator in `YaruPortraitLayout`, `YaruLandscapeLayout` needs to be stateful to manage pane resizing)

jpnurmi (approved):
Ok, sounds good! LGTM :+1: 

Thanks for working on this and sorry it got more laborious than anticipated... :)

PR #417, opened 2022-11-25
YaruCheck/RadioButton: remove extra padding

jpnurmi:
This was probably done to ensure that the focus/hover highlight is inside the widget, but having empty padding around makes it hard to have a compact layout like in the accessibility settings in the settings app.

| Before | After |
|---|---|
| Screenshot from 2022-11-25 07-39-40 | Screenshot from 2022-11-25 07-39-30 |

After this, it will be possible to use  `YaruCheckButton` and `YaruRadioButton` to create pixel-perfect replacements of the checkbox and radio "rows" in the settings app. If `YaruCheckButton` and `YaruRadioButton` had hardcoded empty padding around, the layout would be significantly less dense.

| Before | After |
|---|---|
| image | image |


Jupi007 (approved):
LGFM 👍

The yaru controls already have padding for the focus highlighting, so it is not needed here.

PR #420, opened 2022-11-26
Fix "paneWidth" typo

larsb24:


lib/src/layouts/yaru_master_detail_page.dart
```diff
@@ -145,7 +145,7 @@ class _YaruMasterDetailPageState extends State<YaruMasterDetailPage> {
             onSelected: _setIndex,
             layoutDelegate: widget.layoutDelegate,
             previousPaneWidth: _previousPaneWidth,
-            onLeftPaneWidthChange: (panWidth) => _previousPaneWidth = panWidth,
+            onLeftPaneWidthChange: (paneWidth) => _previousPaneWidth = paneWidth,
```
jpnurmi, line 148: thanks! how about just `width` to avoid that dartfmt wraps the line? :)

```suggestion
            onLeftPaneWidthChange: (width) => _previousPaneWidth = width,
```

PR #430, opened 2022-11-29
Extract YaruWatermark from YaruBanner

jpnurmi:
### Before:
```dart
YaruBanner(
  watermarkIcon: ...
  ...
),
```
### After:
```dart
YaruWatermark(
  watermark: ...
  child: YaruBannerOrAnythingElse(
    ...
  ),
),
```
Screenshot from 2022-11-29 23-01-16

### DIFF
--- a/example/lib/pages/banner_page.dart
+++ b/example/lib/pages/banner_page.dart
@@ -18,25 +18,27 @@ class BannerPage extends StatelessWidget {
       ),
       children: [
         for (int i = 0; i < 20; i++)
-          YaruBanner(
-            title: Text('YaruBanner $i'),
-            subtitle: Column(
-              crossAxisAlignment: CrossAxisAlignment.start,
-              children: [
-                const Text('Description'),
-                if (i.isEven) const Text('Third line'),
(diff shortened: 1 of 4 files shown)


Feichtmeier:
? why?

jpnurmi:
After this, we'd be only a few steps away from
```dart
YaruBanner(
  child: YaruTile(
    ...
  ),
)
```
vs.
```dart
YaruBanner(
  child: Column(
    children: [
      Icon(...),
      Text(...),
    ]
  ),
)
```

jpnurmi:
Frankly speaking, the watermark feature feels a bit store-centric but perhaps there will be other use cases? With this, you can at least control the positioning and translucency. :wink:

Feichtmeier:
Ironically it is currently not used in the store, but I could imagine it can be useful anywhere. Generally it is nice to have a brouder set of components to chose from . If it helps you with the workshop app we can merge this, but generally it is always this configuration vs convient question :P 

Are you sure the stack always has the same size as the child you add to watermark?

Issue #422, opened 2022-11-29
YaruBanner layout


jpnurmi:
I tried using `YaruBanner` for the OS grid in Workshops. I'd like to have the icon on top and the text label below, both horizontally centered. This is not possible with `YaruBanner` because it uses `YaruTile` internally.

Would it make sense to either a) have a configurable banner layout or b) make `YaruBanner` layout-agnostic by allowing any child widget? The latter implies that the watermark feature would also need to be refactored out.

| | Before | After |
|---|---|---|
| Light | ![image](https://user-images.githubusercontent.com/140617/204613607-cd5dd1c9-6dff-47e7-96d1-2687831d8520.png) | ![image](https://user-images.githubusercontent.com/140617/204613397-949b33a0-995b-47c1-a8eb-17f0eb54e4af.png) |
| Dark | ![image](https://user-images.githubusercontent.com/140617/204613541-f6fd10b4-7ccf-4f89-bb9a-7ae83a121340.png) | ![image](https://user-images.githubusercontent.com/140617/204613437-571d574a-dc42-42f2-b5be-4bc5c622d066.png) |

Feichtmeier:
I would prefer a layout parameter also for vertical layout etc. Can also use this in software

jpnurmi:
@Feichtmeier There are two different internal `_Banner` variants used in `YaruBanner` depending on whether `copyIconAsWatermark` is set. Only one of them uses `surfaceTintColor` and they have different elevation values. Are these deliberately chosen values or should it respect `surfaceTintColor` even when no watermark is set? I'm not sure but the elevation value might not make a difference when no surface tint is set. Do we need these different variants at all?

```diff
 _Banner(
   icon: icon,
   iconPadding: iconPadding,
   title: title,
   subtitle: subtitle,
   borderRadius: borderRadius,
-  color: surfaceTintColor ?? defaultCardColor,
-  elevation: light ? 4 : 6,
+  color: defaultCardColor,
+  elevation: light ? 2 : 1,
   mouseCursor: onTap != null ? SystemMouseCursors.click : null,
 ),
```

Feichtmeier:
Yeah 2^2 options
- no water mark, no tint 0 0
- no water mark but tint 0 1
- water mark, no tint 1 0
- water mark and tint 1 1

would be cool to keep those

Feichtmeier:
Edit: sorry. The elevation can be kept to not hide this from the internal card. Also you can control the tint with it

PR #434, opened 2022-11-30
YaruBanner: allow arbitrary child & offer tile for convenience

jpnurmi:
The current tile-based layout is offered via a named `YaruBanner.tile` constructor. It offers the exact same looks as before:

image

The default constructor allows you to pass an arbitrary child widget so you can change the layout as you wish. Here's a simple centered column as a proof of concept:

image

Close: #422

### DIFF
--- a/example/lib/pages/banner_page.dart
+++ b/example/lib/pages/banner_page.dart
@@ -23,7 +23,7 @@ class BannerPage extends StatelessWidget {
               Icons.cloud,
               size: 100,
             ),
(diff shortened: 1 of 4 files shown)


jpnurmi:
Thank you for accepting the `YaruBanner` changes. It made a big difference in Workshop's OS selection!

https://github.com/canonical/workshops/pull/259

Feichtmeier:
Looks great! Thanks for all the work!

PR #455, opened 2022-12-06
Introduce YaruWindowTitleBar & YaruDialogTitleBar

jpnurmi:
`YaruWindowTitleBar` and `YaruDialogTitleBar` are based on `YaruTitleBar` but provide slightly different behavior. The dialog variant has no minimize and maximize buttons by default, doesn't react to double clicks, and calls `Navigator.maybePop` when closed.

| Active | Inactive |
|---|---|
| image | image |
| image | image |


jpnurmi:
Hahaha, needs also some special handling for web...

image

Feichtmeier:
For the window titlebar I would recommend 

`#222222` (Color0xFF222222)
It would be nice to have a color arg though and just make 222222 the default

I would leave the dialogbar as it is

Thank you very much for making this happen!
We could add this to the appbar of the pages on the right side in the xample

Jupi007:
Awesome 😍

Jupi007:
Here is my design point of view on this:

- The title should use: `TextStyle(fontSize: 14, fontWeight: FontWeight.w500)`.
- `AppBar` background should use `Colors.black` with `0.075` of opacity on the light theme, and `0.2` on the dark theme.
- `AppBar` bottom border should use `Colors.black` with `0.1` of opacity on the light theme, and `0.2` on the dark theme.
- Window controls should be spaced of `10` from the window edge, and of `14` from each others.

|  | Light theme | Dark Theme |
| --- | --- | --- |
| `TitleBar` | image | Capture d’écran du 2022-12-06 18-10-36 |
| `DialogBar` | Capture d’écran du 2022-12-06 18-26-35 | Capture d’écran du 2022-12-06 18-27-26 |

Also, I really like the idea of `DialogBar` which can control window position, with the same look as the `TitleBar`. This behaviour "emulate" a second window, while still being in the same one. Also, as the title bar is now flutter, we cannot move the window when a dialog is open, so it is important that dialog can do this action.

Another point, it would be nice to add a backdrop state to the title bar, so we can identify which window is focused more easily.

jpnurmi:
> Here is my design point of view on this

Thank you! Looks very nice!

> Another point, it would be nice to add a backdrop state to the title bar, so we can identify which window is focused more easily.

Is it enough to have different colors for active and inactive states or does it need some sort of backdrop effect?

Feichtmeier:
Do we really always need a background for dialog titlebars? I think even the gnomers are stepping away from this. Maybe we could have an option to remove both border and background color without too many hacks?

Jupi007:
> Is it enough to have different colors for active and inactive states or does it need some sort of backdrop effect?

Maybe can we do like Libadwaita, by remove titlebar background and tone down everything inside?

jpnurmi:
How's this? All of our theme classes can be used as Flutter theme extensions:
```dart
YaruTheme(
  extensions: [
    YaruTitleBarThemeData(
      backgroundColor: MaterialStateProperty.all(Colors.transparent),
    ),
    YaruFooBarThemeData(
      ...
    ),
  ]
  builder: (context, yaru, child) => MaterialApp(
    theme: yaru.theme,
    darkTheme: yaru.darkTheme,
    home: ...
  ),
)
```

jpnurmi:
Any ideas on how to deal with different platforms?

Leave it up to the user to only create a `YaruWindowTitleBar` on Linux? It's pretty straightforward to delete the `GtkHeaderBar` from `my_application.cc` on Linux but on other platforms, you'd have to make the window frameless and then you lose resizing.

jpnurmi:
macOS 🙃

image

jpnurmi:
Both Windows and macOS need some minor tweaks but we might be able to make this work somewhat sensibly on all desktop platforms.

image

Feichtmeier:
> How's this? All of our theme classes can be used as Flutter theme extensions:
> ```dart
> YaruTheme(
>   extensions: [
>     YaruTitleBarThemeData(
>       backgroundColor: MaterialStateProperty.all(Colors.transparent),
>     ),
>     YaruFooBarThemeData(
>       ...
>     ),
>   ]
>   builder: (context, yaru, child) => MaterialApp(
>     theme: yaru.theme,
>     darkTheme: yaru.darkTheme,
>     home: ...
>   ),
> )
> ```

Very nice

Feichtmeier:
> Any ideas on how to deal with different platforms?
> 
> Leave it up to the user to only create a `YaruWindowTitleBar` on Linux? It's pretty straightforward to delete the `GtkHeaderBar` from `my_application.cc` on Linux but on other platforms, you'd have to make the window frameless and then you lose resizing. Oh, and on Web you don't want anything there.

Depends on you 🙂
Currently this lib targets Linux and web
But if you can make it work on Windows and MacOS too why not . The design works on all desktops

Feichtmeier:
>  without touching my_application.cc

That's pretty awesome!

whiskeyPeak:
Wow! This looks absolutely gorgeous!

larsb24:
Looks awesome!

One potential problem I could see though is the expected behaviour when dialogs are opened. I don't know where the intention lies, which is why I'm saying “potential” but not being able to access the titlebar or it's functions while dialogs are open could be unexpected by some users.

jpnurmi:
Thanks. The way it currently behaves is that dialog title bars allow moving the entire window and opening the popup menu. This makes dialogs feel almost like top-level windows. :)

Jupi007:
Yes, and this is a really good idea, because as the headerbar is now a flutter widget, it is hidden when you open a dialog.
Be able to drag the window from the dialog fix this issue.

jpnurmi:
> This PR makes it possible to call `windowManager.setTitleBarStyle(TitleBarStyle.hidden)` to hide the GTK or Handy header bar out of the way on Linux without touching `my_application.cc`.
> 
> * [Linux: make `setTitleBarStyle()` GTK/HDY/CSD/SSD compatible leanflutter/window_manager#249](https://github.com/leanflutter/window_manager/pull/249)
> 
> I used the same WindowManager API when testing Windows and macOS earlier. This could be a nice and cross-platform way to set things up.

I have to eat my words. It's not possible to hide the native title/header bar from Dart early enough. It's visible for a brief moment during startup.

Feichtmeier:
> I have to eat my words. It's not possible to hide the native title/header bar from Dart early enough. It's visible for a brief moment during startup.

does this mean that the version where you do not touch my_application.cc shows the bar for a short moment or is it even shown when you make the known changes from handy_window?

Feichtmeier (approved):
I can not say much for the .cc changes, but the dart changes look good and intuitive to use and I tested this locally and it works

grafik

In the name of all linux desktop users, thank you jp! :+1: :1st_place_medal:

Feichtmeier:
Uh I found a little issue, the clos ebutton dissapeared for the close buton in the xample
grafik

jpnurmi:
I've updated the example. We clearly need some more convenience for overriding the state...

Jupi007:
This looks awesome, but I can't run the example: 😥

Capture d’écran du 2023-01-09 15-08-10

Feichtmeier:
flutter clean ?

Jupi007:
Sadly there is still the same problem even after a `flutter clean`.

jpnurmi:
@Jupi007 What about `flutter pub upgrade`? Does it make any difference? Also, which yaru_widgets.dart commit are you on, and which Flutter version are you using?

Jupi007:
@jpnurmi I had already tried `pub upgrade`, without success.

My flutter version is:

```
[✓] Flutter (Channel stable, 3.3.10, on Ubuntu 22.10 5.19.0-28-generic, locale fr_FR.UTF-8)
```

jpnurmi:
Testing `YaruWindowTitleBar` with Workshops :+1: 

image

Feichtmeier:
Wonder why it worked for me :thinking: 
Maybe just delete the local repo and clone again?

Jupi007:
> Wonder why it worked for me :thinking: 
> Maybe just delete the local repo and clone again?

Will try this once back home 👍

Feichtmeier:
then the only difference that comes to my mind that is left is maybe the locale? try running flutter run with en_US

Jupi007:
Found the problem: it is broken in debug mode.
It runs fine with `flutter run`.

I can perfectly run the example app in both mode from `main`, it is only this PR who is broken in debug.

jpnurmi:
Gotcha, so
```dart
isMovable().catchError((_) => true),
```
doesn't actually end up catching the error because `WindowManager.isMovable()` internally awaits the method channel invocation...

jpnurmi:
@Jupi007 Ok, should be fixed now. Please give it a try when you have time. :pray:

Jupi007:
Works 🥳
This is really awesome, thank you so much 😃

Jupi007 (approved):
LGFM 🚀

jpnurmi:
is that yaru_widgets 2.0?

Feichtmeier:
Yes 🖐️👍 I'll do a release tomorrow if you aren't faster 😉 thank you JP

PR #464, opened 2023-01-01
YaruTitleBar: make themable and use window controls

jpnurmi:
Visual changes extracted from #455, leaving out any actual top-level window-related functionality i.e. `YaruWindowTitleBar`, `YaruDialogTitleBar`, and `YaruWindowController`.

NOTE: This is a breaking change. `YaruTitleBar` no longer creates a `YaruCloseButton` that calls `Navigator.pop`. Instead, it has a set of properties that can be used to manage which `YaruWindowControls` to create.

P.S. `YaruWindowTitleBar` and `YaruDialogTitleBar` are still coming but without `window_manager` dependency as originally proposed in #455. :)

### DIFF
--- a/lib/src/controls/yaru_title_bar.dart
+++ b/lib/src/controls/yaru_title_bar.dart
@@ -1,7 +1,10 @@
 import 'package:flutter/material.dart';
 
 import '../constants.dart';
-import 'yaru_close_button.dart';
+import 'yaru_title_bar_theme.dart';
+import 'yaru_window_control.dart';
+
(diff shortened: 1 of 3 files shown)


Feichtmeier:
Cool, could you add some .close() constructor that adds a close button by default?

jpnurmi:
There's a separate `YaruDialogTitleBar` in the works that will call `Navigator.pop` out of the box. With the "plain" `YaruTitleBar` you can do:
```dart
YaruTitleBar(
  isClosable: true,
  onClose: Navigator.of(context).pop,
)
```

PR #468, opened 2023-01-07
MD: fix page controller updates

jpnurmi:
`MasterDetailPage` can be rebuilt with either a) a different `controller`, or b) a different `length` and `initialIndex`. Make sure `didUpdateWidget` checks for both cases.

Ref: ubuntu-flutter-community/settings#421


Feichtmeier (approved):
Thank you! Maybe you were right after all and we should just add a required controller some day 🤠

jpnurmi:
It was the opposite, actually. :see_no_evil: A controller was required in the beginning because it made our internal implementation simple. However, it was inconvenient to use because often you don't really need one. Therefore I proposed making it easier for the user by allowing them to pass an initial index and length just like they did before, at the cost of making our internal implementation more complex...

PR #473, opened 2023-01-09
Navigation/MD pages: pop all but first on select

d-loose:
Clears the navigator stack of the right pane in `YaruLandScapeLayout` and `YaruNavigationPage` when (re-)selecting an entry in the left pane.

See https://github.com/ubuntu-flutter-community/software/issues/254

Edit: This requires us to use the navigator provided by the respective Yaru page, rather than nesting another one below.

## Pull request checklist

    | |Before|After|
    |-|-|-|
    |Light| | |
    |Dark| | |

### DIFF
--- a/lib/src/layouts/yaru_landscape_layout.dart
+++ b/lib/src/layouts/yaru_landscape_layout.dart
@@ -42,6 +42,7 @@ const _kLeftPaneResizingRegionAnimationDuration = Duration(milliseconds: 250);
 
 class _YaruLandscapeLayoutState extends State<YaruLandscapeLayout> {
   late int _selectedIndex;
+  final _navigatorKey = GlobalKey<NavigatorState>();
 
   double? _paneWidth;
   double? _initialPaneWidth;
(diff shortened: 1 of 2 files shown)


Feichtmeier:
Thanks @d-loose !

The the yaru widgets example does not have a nested navigation I tried your fork and branch in the software app and it does not work for me after clean, upgrade and get

> Edit: This requires us to use the navigator provided by the respective Yaru page, rather than nesting another one below.

Ah! So we need a second change in for example software?

d-loose:
> The the yaru widgets example does not have a nested navigation

I'll add an example page and a test.

> I tried your fork and branch in the software app and it does not work for me after clean, upgrade and get
> > Edit: This requires us to use the navigator provided by the respective Yaru page, rather than nesting another one below.
> 
> Ah! So we need a second change in for example software?

Yes - with this approach we only have access to the navigator provided by the Yaru page itself. But for the software store that's not really a problem, since all the navigators there only contain a single page (I'm not even sure why they're there, probably from a time when the yaru pages looked different :D)

Feichtmeier:
@d-loose ok, thanks !

PR #475, opened 2023-01-09
Improve popupmenubutton style consistency

whiskeyPeak:
This PR matches the border radius and colour of the popupbutton to the popupmenu.

## Pull request checklist

    | |Before|After|
    |-|-|-|
    |Light|Screenshot from 2023-01-09 23-40-51 | Screenshot from 2023-01-09 23-40-07|
    |Dark| Screenshot from 2023-01-09 23-50-59| Screenshot from 2023-01-09 23-50-06|


whiskeyPeak:
I tried to expose the ```shape``` property as well, but because ```Theme.of(context)``` can't be const, that wasn't possible.

Feichtmeier:
hmmm do we want all popup menus to have this border radius? then this is a change for yaru.dart
or do we only want this for this specific "fake dropdowns" ?

CC @jpnurmi @Jupi007

Jupi007:
Imo, this should be applied to all popup menus.

Feichtmeier:
> Imo, this should be applied to all popup menus.

Then we should change this in yaru.dart and if I am not wrong this change is then no longer needer :thinking:

Jupi007:
> Then we should change this in yaru.dart and if I am not wrong this change is then no longer needer thinking

This is also my opinion.

whiskeyPeak:
Sounds good! I'll open up a PR over there then.

Feichtmeier:
@whiskeyPeak thank you!

While you are at it, could you instead of changing the border to match the button border change the button border to match the outlined button border? =) Wanted to do this today but since you are on it would be nice

whiskeyPeak:
> @whiskeyPeak thank you!
> 
> While you are at it, could you instead of changing the border to match the button border change the button border to match the outlined button border? =) Wanted to do this today but since you are on it would be nice

The problem with that is that every other yaru widget uses the same border colour as the button border. ie: YaruOptionButtons, YaruSection, YaruBanner, YaruCarousel...
So if we want to change the border colour of the PopupMenuButton, it would probably be best if we do it via modifying the constant.

PR #482, opened 2023-01-10
YaruTitleBar: consider active by default

jpnurmi:
Instead of requiring an explicit `isActive: true`, check that it's not explicitly `false`. In other words, consider `null` active, too. This helps to render dialogs correctly when they become visible. It might take a noticeable amount of time to fetch the information over the platform channel.


jpnurmi:
The issue is barely visible in this screencast. Look at the close button very closely when navigating forward. It turns disabled for a moment.

[Screencast from 2023-01-10 17-52-52.webm](https://user-images.githubusercontent.com/140617/211819933-f8464a52-fb32-4772-b240-889adc1f0adb.webm)

Feichtmeier (approved):
https://user-images.githubusercontent.com/15329494/211820498-6d37a73f-78aa-4cfb-84e3-ec149c182d79.mp4

Looks like it's fixed if the main navigation and the nested naviation in the example are enough to prove?

Feichtmeier:
oh, you have forgotten the goldens :medal_sports: :smile:  @jpnurmi

PR #485, opened 2023-01-11
Fix the example to respect the system theme variant

### LINKED ISSUE #477: Question about how to change YaruRadio color?
Can somebody tell me how to change YaruRadio color? when launch in my computer, my color which is blue is different from demo which is orange.

jpnurmi:
It should default to the system theme variant on startup and follow it until explicitly selected on the color disk page.

Closes #477


Feichtmeier:
grafik

Checked out -> works :+1:

Issue #483, opened 2023-01-10
Title bar constants


jpnurmi:
Could we expose the necessary window control and title bar constants to make it possible to calculate how much space they eat up?

![image](https://user-images.githubusercontent.com/140617/211677869-43e2099f-b61a-4a27-8ec5-ffeb564157ef.png)

[Screencast from 2023-01-10 23-36-58.webm](https://user-images.githubusercontent.com/140617/211677459-35e1e18f-b4a5-43a0-92aa-2d444dd2e883.webm)

PR #486, opened 2023-01-11
Expose title bar / window control metrics

jpnurmi:
Makes it possible to calculate how much space the window controls take for alignment purposes. No visual changes.

Close: #483


Feichtmeier (approved):
grafik
checked out -> works/cant find visual regressions

PR #490, opened 2023-01-11
Fix stretched yaru togglables

### LINKED ISSUE #484: Checkbox, Radio, and Switch stretch
image

```dart
import 'package:yaru/yaru.dart';
import 'package:yaru_widgets/yaru_widgets.dart';

void main() {
  runApp(
    YaruTheme(
      builder: (context, yaru, child) => MaterialApp(
        theme: yaru.theme,
        darkTheme: yaru.darkTheme,
        home: const MyHomePage(),
      ),
    ),
  );
}

class MyHomePage extends StatelessWidget {
  const MyHomePage({super.key});

  @override
  Widget build(BuildContext context) {
    return Scaffold(
      body: Padding(
        padding: const EdgeInsets.all(kYaruPagePadding),
        child: Table(
          border: TableBorder.all(color: Theme.of(context).dividerColor),
          children: [
            const TableRow(
              children: [
                SizedBox.shrink(),
                Text('Material'),
                Text('Yaru'),
              ],
            ),
            TableRow(
              children: [
                const Text('Checkbox'),
                Checkbox(value: true, onChanged: (_) {}),
                YaruCheckbox(value: true, onChanged: (_) {}),
              ],
            ),
            TableRow(
              children: [
                const Text('Radio'),
                Radio(value: true, groupValue: true, onChanged: (_) {}),
                YaruRadio(value: true, groupValue: true, onChanged: (_) {}),
              ],
            ),
            TableRow(
              children: [
                const Text('Switch'),
                Switch(value: true, onChanged: (_) {}),
                YaruSwitch(value: true, onChanged: (_) {}),
              ],
            ),
          ],
        ),
      ),
    );
  }
}
```

Jupi007:
- Center the painter to fix the stretch effect ;
- Add `RepaintBoundary` widget around the painter to keep a sharp look ;
- Correct weird logic of putting a `Padding` in a `SizedBox`, which was require to add up padding value to the dimensions.

@jpnurmi as you can see, I'm now using a personal fork ;)

**Before:**

Capture d’écran du 2023-01-11 11-24-23

**After:**

Capture d’écran du 2023-01-11 11-19-47

## Pull request checklist

Fixes #484

### DIFF
--- a/lib/src/controls/yaru_togglable.dart
+++ b/lib/src/controls/yaru_togglable.dart
@@ -250,33 +250,38 @@ abstract class YaruTogglableState<S extends YaruTogglable> extends State<S>
 
     return _buildSemantics(
       child: _buildEventDetectors(
-        child: SizedBox(
-          width: togglableSize.width + activableAreaPadding.horizontal,
-          height: togglableSize.height + activableAreaPadding.vertical,
-          child: Padding(
-            padding: activableAreaPadding,
-            child: CustomPaint(
-              size: togglableSize,
-              painter: painter
-                ..interactive = widget.interactive
-                ..hover = _hover
-                ..focus = _focus
-                ..active = _active
-                ..checked = widget.checked
-                ..oldChecked = _oldChecked
-                ..position = _position
-                ..sizePosition = _sizePosition
-                ..indicatorPosition = _indicatorPosition
-                ..uncheckedColor = uncheckedColor
-                ..uncheckedBorderColor = uncheckedBorderColor
-                ..checkedColor = checkedColor
-                ..checkmarkColor = checkmarkColor
-                ..disabledUncheckedColor = uncheckedDisabledColor
-                ..disabledUncheckedBorderColor = uncheckedDisabledBorderColor
-                ..disabledCheckedColor = checkedDisabledColor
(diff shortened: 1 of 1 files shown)


jpnurmi (approved):
Thanks, LGTM! :+1: 

Is the `RepaintBoundary` absolutely necessary? The Material toggleables don't seem to use one either. I git a bit cautious because I debugged earlier what it does under the hood...

jpnurmi:
> Well yes, because the togglable becomes blurry in the horizontal direction without one.

Does it happen with the Material togglables too?

> And I sadly don't know any other way to fix that problem...

I think it's more like a coincidence that `RepaintBoundary` force-aligns to pixels because it creates a separate `GtkWidget` that cannot have floating point coordinates. It's just that using this trick for every checkbox is almost like creating floating GTK checkboxes on top of the Flutter view. :grinning:

Jupi007:
> Does it happen with the Material togglables too?

Well no, and I don't understand why, but material togglables are not affected by the pixel alignment problem.

> I think it's more like a coincidence that RepaintBoundary force-aligns to pixels because it creates a separate GtkWidget that cannot have floating point coordinates. It's just that using this trick for every checkbox is almost like creating floating GTK checkboxes on top of the Flutter view. grinning

I will try to find another solution then.

Issue #489, opened 2023-01-11
Clean up and document constants


jpnurmi:
- prefix all public constants (`kYaruFoo`)
- document all public constants
- hide constants that are shared for internal implementation

PR #494, opened 2023-01-11
Hide `kTogglableActiveResizeFactor` out of the public API

jpnurmi:
@Jupi007 is this ok to hide? do we need it in the public `constants.dart`? it's only used in `togglable.dart`.

Ref: #489


Jupi007 (approved):
It is completely okay to move this 👍
I surely put it here because I was access to it from both checkbox and radio fiels in early versions. Anyway, it's place is here you moved it 🙂

PR #495, opened 2023-01-11
Move `kCheckboxRadio*` constants out of the public API

jpnurmi:
@Jupi007 What about this? I left it to a separate PR because this could be controversial...

Ref: #489


Jupi007 (approved):
Thank you for asking, imo it is fine to move them like you did.
Those two bicephal public const has always been annoying me a bit :D

jpnurmi:
Phew, glad to hear that so we can get rid of them in auto-completion and docs :)
https://pub.dev/documentation/yaru_widgets/2.0.0-beta-3/yaru_widgets/yaru_widgets-library.html#constants

For a while, I did consider introducing a separate file for these but meh... maybe we can make these themable later, though.

PR #474, opened 2023-01-09
add NavigationPage

d-loose:
https://user-images.githubusercontent.com/113362648/211363557-12af5c03-35bf-462d-9b08-c4c7d44551b9.mp4

I'll add some widget tests in a separate PR

## Pull request checklist

- [x] This PR does not introduce visual changes, **or**
  - I ran `flutter test --update-goldens` and committed the changes if there were any, **or**
  - I added before/after/light/dark screenshots if the visual changes I made were not covered by golden tests.
    | |Before|After|
    |-|-|-|
    |Light| | |
    |Dark| | |

PR #500, opened 2023-01-12
Example: fix navigation page title

jpnurmi:
| Before | After |
|---|---|
| image | image |
| Screenshot from 2023-01-12 09-08-29 | image |


d-loose:
Sorry, didn't realize I only changed the tooltip in #474 :D

jpnurmi:
No worries! We should probably clean this up a bit to avoid having to define the same string twice... There used to be pages with custom builders so the title didn't always match with the tooltip.

jpnurmi:
- https://github.com/ubuntu/yaru_widgets.dart/pull/504

PR #513, opened 2023-01-13
Example: use CSD with yaruwindowtitlebar if not web

Feichtmeier:
Example

Desktop
grafik

Web
grafik

## Pull request checklist

- [X] This PR does not introduce visual changes, **or**
  - I ran `flutter test --update-goldens` and committed the changes if there were any, **or**
  - I added before/after/light/dark screenshots if the visual changes I made were not covered by golden tests.
    | |Before|After|
    |-|-|-|
    |Light| | |
    |Dark| | |

PR #514, opened 2023-01-13
YaruWindowTitleBar: improve defaults for web

jpnurmi:
Now that we have the "undecorated" style, we can use that on web so YaruWindowTitleBar becomes a drop-in replacement for AppBar.


Feichtmeier:
Checked it out then clean then run... :thinking: 

Is this a superseeding PR for #513 ?

Because if it is it might not work as I would expect it since I still have the double titlebar:
grafik

jpnurmi:
This allows you to replace `kIsWeb ? AppBar(...) : YaruWindowTitleBar(...)` in #513 with just `YaruWindowTitleBar(...)`.

Issue #515, opened 2023-01-13
YaruTitleBar trailing vs. actions


jpnurmi:
Should we change `Yaru*TitleBar` to allow passing in a list of `actions` instead of a single `trailing` widget? This would make it more straightforward to replace `AppBar` with `Yaru*TitleBar`. What do you think?

Feichtmeier:
Yes good idea

PR #517, opened 2023-01-14
YaruTitleBar: trailing -> actions

jpnurmi:
To make `YaruTitleBar` closer to a drop-in replacement of `AppBar`.

Close: #515

### DIFF
--- a/lib/src/controls/yaru_title_bar.dart
+++ b/lib/src/controls/yaru_title_bar.dart
@@ -18,7 +18,7 @@ class YaruTitleBar extends StatelessWidget implements PreferredSizeWidget {
     super.key,
     this.leading,
     this.title,
-    this.trailing,
+    this.actions,
     this.centerTitle,
     this.titleSpacing,
     this.foregroundColor,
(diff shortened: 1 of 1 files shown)


lib/src/controls/yaru_title_bar.dart
```diff
@@ -207,7 +207,7 @@ class YaruTitleBar extends StatelessWidget implements PreferredSizeWidget {
               Row(
                 mainAxisSize: MainAxisSize.min,
                 children: [
-                  if (trailing != null) trailing!,
+                  ...?actions,
```
Feichtmeier, line 210: this shortcut is so cool, didn't know this :heart_eyes_cat: 
How I can I ever love Java again :sob:

Feichtmeier (approved):

jpnurmi:
Hmm, I noticed that `AppBar` has an interesting check for determining whether to center the title depending on the amount of actions. I wonder if we should have something similar... :thinking: 

```dart
  bool _getEffectiveCenterTitle(ThemeData theme) {
    bool platformCenter() {
      assert(theme.platform != null);
      switch (theme.platform) {
        case TargetPlatform.android:
        case TargetPlatform.fuchsia:
        case TargetPlatform.linux:
        case TargetPlatform.windows:
          return false;
        case TargetPlatform.iOS:
        case TargetPlatform.macOS:
          return actions == null || actions!.length < 2;
      }
    }

    return centerTitle
      ?? theme.appBarTheme.centerTitle
      ?? platformCenter();
  }
```

Feichtmeier:
Still we mimic the "api" here in most places: sounds good

jpnurmi:
Sure, it was just something that caught my attention while checking how AppBar handles actions. I think this change is fine regardless of what we decide to do with centerTitle.

PR #519, opened 2023-01-14
Example: move code snippet button out of the app bar

jpnurmi:
It takes too much space when we switch to a window title bar. How about using a FAB instead?

| Before | After |
|---|---|
| image | Screenshot from 2023-01-14 17-25-26 |

### DIFF
--- a/example/lib/code_snippet_button.dart
+++ b/example/lib/code_snippet_button.dart
@@ -35,6 +35,9 @@ class CodeSnippedButton extends StatelessWidget {
       ),
       child: const Icon(YaruIcons.code),
       tooltip: 'Example snippet',
+      foregroundColor: Theme.of(context).colorScheme.onSurface,
+      backgroundColor: PopupMenuTheme.of(context).color,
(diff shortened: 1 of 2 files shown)


jpnurmi:
This is how it would look like if we kept it in the window title bar, which is why I want to move it...

Screenshot from 2023-01-14 17-15-22

Feichtmeier:
FAB is nice! We finally have a FAB :dancers:

Jupi007:
I like it 👌 but isn't this green a little bit too much attention grabbing?

jpnurmi:
Agreed! Any suggestion what color to use?

Feichtmeier:
the color from the popup menus :)

jpnurmi:
```dart
foregroundColor: Theme.of(context).colorScheme.onSurface,
backgroundColor: PopupMenuTheme.of(context).color,
shape: PopupMenuTheme.of(context).shape,
```

| Normal | Hovered |
|---|---|
| image |image |

Jupi007:
@jpnurmi much better :+1:

PR #520, opened 2023-01-14
Remove all RepaintBoundary widgets

### LINKED ISSUE #492: Controls: `RepaintBoundary` absolutely necessary
Thanks, LGTM! :+1: 

Is the `RepaintBoundary` absolutely necessary? The Material toggleables don't seem to use one either. I git a bit cautious because I debugged earlier what it does under the hood...

_Originally posted by @jpnurmi in https://github.com/ubuntu/yaru_widgets.dart/pull/490#pullrequestreview-1243722665_

Jupi007:
This PR removes all RepaintBoundary widgets, which can cause performance issue.
The only side effect is that window controls and togglable widgets will no more be aligned on physical display. It will result in blurred widgets in some situations, like inside alignment widgets.

I taken an attentive look, and all material widgets got the same behavior of blurred look.

Side by side, material and yaru checkboxes:

[Capture vidéo du 2023-01-14 17-39-38.webm](https://user-images.githubusercontent.com/36476595/212485708-e4c80e80-9710-4ce7-ac2d-bb6193528edf.webm)

Fixes #492

## Pull request checklist

### DIFF
--- a/lib/src/controls/yaru_togglable.dart
+++ b/lib/src/controls/yaru_togglable.dart
@@ -257,31 +257,28 @@ abstract class YaruTogglableState<S extends YaruTogglable> extends State<S>
             width: togglableSize.width,
             height: togglableSize.height,
             child: Center(
-              child: RepaintBoundary(
-                child: CustomPaint(
-                  size: togglableSize,
-                  painter: painter
-                    ..interactive = widget.interactive
-                    ..hover = _hover
-                    ..focus = _focus
-                    ..active = _active
-                    ..checked = widget.checked
-                    ..oldChecked = _oldChecked
-                    ..position = _position
-                    ..sizePosition = _sizePosition
-                    ..indicatorPosition = _indicatorPosition
-                    ..uncheckedColor = uncheckedColor
-                    ..uncheckedBorderColor = uncheckedBorderColor
-                    ..checkedColor = checkedColor
-                    ..checkmarkColor = checkmarkColor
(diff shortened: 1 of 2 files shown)


Feichtmeier:
> like inside alignment widgets

So every time we use a checkbox within Alignment it is blurred?

Jupi007:
Yes, half the time it will be blurred depending on the window size.
Ex: you have a 701px width window, using a `Center` widget, will move a 20px width checkbox at 340.5px from the left side, which will look blurred as the borders will be drawn on two physical pixels.

jpnurmi:
I think it's kind of a fundamental issue in Flutter that there's no sensible way to control pixel alignment. It's not really an issue elsewhere besides on lower-DPI monitors typically used on desktops, though...

jpnurmi:
For example, when centering stuff in Qt Quick: https://doc.qt.io/qt-5/qml-qtquick-item.html#anchors.alignWhenCentered-prop
And you can also control whether scrolling always lands on pixels https://doc.qt.io/qt-6/qml-qtquick-flickable.html#pixelAligned-prop

Jupi007:
Yeah, this is not that a question of low dpi, but more a problem of the way that flutter aligns widgets on subpixels, even at 100% scale.

jpnurmi (approved):
LGTM :+1:

Feichtmeier (approved):

Jupi007:
Will do the same thing in yaru_icons

PR #521, opened 2023-01-14
Example: make better use of YaruWindowTitleBar

jpnurmi:
image
image
image
image

Supercedes #513

### DIFF
--- a/example/lib/example.dart
+++ b/example/lib/example.dart
@@ -35,58 +35,76 @@ class _ExampleState extends State<Example> {
   Widget build(BuildContext context) {
     final model = context.watch<ExampleModel>();
 
-    final pageItems = examplePageItems;
     return model.compactMode
-        ? const _CompactPage()
-        : YaruMasterDetailPage(
-            layoutDelegate: const YaruMasterResizablePaneDelegate(
-              initialPaneWidth: 280,
-              minPageWidth: kYaruMasterDetailBreakpoint / 2,
-              minPaneWidth: 175,
-            ),
-            length: pageItems.length,
-            tileBuilder: (context, index, selected) => YaruMasterTile(
-              leading: pageItems[index].iconBuilder(context, selected),
-              title: buildTitle(context, pageItems[index]),
-            ),
-            pageBuilder: (context, index) => YaruDetailPage(
-              appBar: AppBar(
-                leading: Navigator.of(context).canPop()
-                    ? const YaruBackButton()
-                    : null,
-                title: buildTitle(context, pageItems[index]),
-              ),
-              body: pageItems[index].pageBuilder(context),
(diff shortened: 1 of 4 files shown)


jpnurmi:
There's one special page that doesn't play nicely:

image
image

Feichtmeier:
Yes it has an AppBar instead of a yaru bar

https://github.com/ubuntu/yaru_widgets.dart/blob/main/example/lib/pages/navigation_page.dart#L26

jpnurmi:
Damn, the idea of using a divided title bar with master-detail and shared one navi-rail doesn't work then. 🤔

jpnurmi:
What if we make the nested pages use YaruWindowTitleBar and filter out the whole page item from navi-rail?

| MD | NR |
|---|---|
| image | image |

Feichtmeier:
> Damn, the idea of using a divided title bar with master-detail and shared one navi-rail doesn't work then. thinking

Well we had a version of the layouts which needed those titlebuilders for the appbar in the scaffolds. Pro was that we did not use so many scaffolds then. But if we would do this again we would need one more builder for the appbar title because we are using the titlebuilders for the master tiles and navi items now

Feichtmeier:
> What if we make the nested pages use YaruWindowTitleBar and filter out the whole page item from navi-rail?

I didn't understand this :thinking: could you elaborate ?

jpnurmi:
Proposal:
```dart
  PageItem(
    title: 'YaruNavigationPage',
    iconBuilder: (context, selected) => selected
        ? const Icon(YaruIcons.compass_filled)
        : const Icon(YaruIcons.compass),
    pageBuilder: (_) => const NavigationPage(),
    supportedLayouts: {YaruMasterDetailPage}, // <=== added (defaults to {YaruMasterDetailPage,YaruNavigationPage})
  ),
```

Feichtmeier:
> Proposal:
> 
> ```dart
>   PageItem(
>     title: 'YaruNavigationPage',
>     iconBuilder: (context, selected) => selected
>         ? const Icon(YaruIcons.compass_filled)
>         : const Icon(YaruIcons.compass),
>     pageBuilder: (_) => const NavigationPage(),
>     supportedLayouts: {YaruMasterDetailPage}, // <=== added (defaults to {YaruMasterDetailPage,YaruNavigationPage})
>   ),
> ```

ahhhh! :) 

Yeah maybe this is a good idea. Eventually we could add more layouts in the future with fixed appbars which require title widgets plus nullable leading widgets

jpnurmi:
> Eventually we could add more layouts in the future with fixed appbars which require title widgets plus nullable leading widgets

Yaru wizard would be cool. :) I've learned with the installer that wizards can have awfully complicated conditional navigation rules, though.

Issue #516, opened 2023-01-13
Delay when pressing the minimise, maximise and close buttons.


whiskeyPeak:
This is kinda hard to show on video, but there is a small delay after pressing those buttons before the window is finely minimised, maximised or closed.

I think this may be because the onDoubleTap argument is used, so Flutter has to wait for a given time to check for a 2nd tap event to occur.

For example, when double pressing the minimise button, the window will maximise. I checked this behaviour with other apps and they minimise/maximise/close as soon as the first tapDown event is completed.

Edit:

This also applies to the tab buttons in the updates page of the software store but doesn't apply to the text-field in the explore and installed pages.

jpnurmi:
Hah, how did I not notice that! This is indeed due to how gesture negotiation works in Flutter. A single tap gesture can only declare itself as a winner after waiting for 300ms when there's a double-tap gesture in the same arena... :(

Feichtmeier:
It's not perfect but I don't think this should be a show-stopper for CSD

whiskeyPeak:
We could replace the gesture detector widget with a Listener widget for those specific buttons.

```
import 'package:flutter/material.dart';

void main() {
  runApp(const MyApp());
}

class MyApp extends StatelessWidget {
  const MyApp({Key? key}) : super(key: key);

  @override
  Widget build(BuildContext context) {
    return const MaterialApp(
      home: HomePage(),
    );
  }
}

class HomePage extends StatefulWidget {
  const HomePage({Key? key}) : super(key: key);

  @override
  State<HomePage> createState() => _HomePageState();
}

class _HomePageState extends State<HomePage> {
  @override
  Widget build(BuildContext context) {
    return Scaffold(
      body: GestureDetector(
        onTap: () {
          print("tap from 1st gesture detector");
        },
        onDoubleTap: () {
          print("double tap from 1st gesture detector");
        },
        child: Container(
          width: 500,
          height: 500,
          color: Colors.grey,
          child: Center(
            child: Listener(
              onPointerUp: (PointerUpEvent event) {
                print("tap from the listener");
              },
              child: const Text("Hello there"),
            ),
          ),
        ),
      ),
    );
  }
}

```
Definitely feels faster imo.

jpnurmi:
This affects any widgets users may put in the title bar. Also, this makes me realize that the behavior with movable tabs feeling a bit off in the Workshops app might have something to do with this too. Thanks for bringing it up! 👍

[...]

PR #524, opened 2023-01-14
YaruTitleBar: fix single-click delay

jpnurmi:
The problem was that `GestureDetector.onDoubleTap` creates an "eager" double-tap recognizer that forces a 300ms delay for any (single-)tap recognizer in the gesture arena.

This PR drops `GestureDetector` in favor of `RawGestureDetector` that allows using custom gesture recognizers, and implements a "passive" double-tap recognizer that is just an extended (single-)tap recognizer but detects double-taps from incoming (single-)taps. It keeps a history of the previous tap and compares the duration and distance with help of the constants defined in `flutter/gestures.dart`.

Fixes: #516

### DIFF
--- a/lib/src/controls/yaru_title_bar.dart
+++ b/lib/src/controls/yaru_title_bar.dart
@@ -5,6 +5,7 @@ import 'package:flutter/material.dart';
 import 'package:yaru_colors/yaru_colors.dart';
 
 import '../constants.dart';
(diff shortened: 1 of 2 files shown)


Jupi007 (approved):
LGFM :+1: 

Thanks very much @jpnurmi 💪‍
This feels so much better without delay!

Issue #518, opened 2023-01-14
YaruTitleBar: null check operator used on a null value


jpnurmi:
`AppBarTheme.titleTextStyle` is null by default (when not using the Yaru theme).

```dart
import 'package:flutter/material.dart';
import 'package:yaru_widgets/yaru_widgets.dart';

void main() {
  runApp(
    const MaterialApp(
      home: Scaffold(
        appBar: YaruWindowTitleBar(),
      ),
    ),
  );
}
```

```
Launching lib/main.dart on Linux in debug mode...
Connecting to VM Service at ws://127.0.0.1:44129/YaaNpS86WYE=/ws

════════ Exception caught by widgets library ═══════════════════════════════════
The following _CastError was thrown building YaruTitleBar(dirty, dependencies: [_InheritedTheme, _LocalizationsScope-[GlobalKey#eff3e]]):
Null check operator used on a null value

The relevant error-causing widget was
YaruTitleBar
When the exception was thrown, this was the stack
#0      YaruTitleBar.build
#1      StatelessElement.build
#2      ComponentElement.performRebuild
#3      Element.rebuild
#4      ComponentElement._firstBuild
#5      ComponentElement.mount
...     Normal element mounting (36 frames)
#41     Element.inflateWidget
#42     MultiChildRenderObjectElement.inflateWidget
#43     MultiChildRenderObjectElement.mount
...     Normal element mounting (379 frames)
#422    Element.inflateWidget
#423    MultiChildRenderObjectElement.inflateWidget
#424    MultiChildRenderObjectElement.mount
...     Normal element mounting (416 frames)
#840    Element.inflateWidget
#841    Element.updateChild
#842    RenderObjectToWidgetElement._rebuild
#843    RenderObjectToWidgetElement.mount
#844    RenderObjectToWidgetAdapter.attachToRenderTree.<anonymous closure>
#845    BuildOwner.buildScope
#846    RenderObjectToWidgetAdapter.attachToRenderTree
#847    WidgetsBinding.attachRootWidget
#848    WidgetsBinding.scheduleAttachRootWidget.<anonymous closure>
(elided 4 frames from class _RawReceivePortImpl, class _Timer, and dart:async-patch)
════════════════════════════════════════════════════════════════════════════════

════════ Exception caught by widgets library ═══════════════════════════════════
Null check operator used on a null value
The relevant error-causing widget was
YaruTitleBar
════════════════════════════════════════════════════════════════════════════════
```

[...]

PR #535, opened 2023-01-16
YaruTitleBar: use TextTheme.titleLarge as a fallback

jpnurmi:
When AppBarTheme.titleTextStyle is null.

Close: #518


jpnurmi:
Screenshot from 2023-01-16 20-27-39

PR #538, opened 2023-01-18
Update goldens

jpnurmi:
Not sure where the sudden difference comes from but let's try what the CI says...


Jupi007:
Thanks, you was quicker than me :D
Looks like the diff comes from the titlebar height.

jpnurmi:
Ah, good catch. Makes sense

Jupi007 (approved):
Okay, thanks.
LGFM so 👍

PR #539, opened 2023-01-18
YaruPopupMenuButton: sync border with OutlinedButton

jpnurmi:
| Before | After |
|---|---|
| Screenshot from 2023-01-18 22-46-48 | Screenshot from 2023-01-18 22-46-35 |
| image | image |

Ref: ubuntu/yaru.dart#244


whiskeyPeak:
The border radius seems to be slightly different between the 2 of them.

PR #552, opened 2023-01-20
Example: fix the settings bottom bar look sensible with Flutter beta

jpnurmi:
M3 bottom bar has gained default padding and a height of 80px on the beta channel. We can switch to a simpler structure to avoid having to customize it in the example. All we need is a tile with an opaque background and a little padding around. :)

| | Before | After |
|---|---|---|
| Stable | image | image |
| Beta | image | image |

NOTE: The white divider is a separate issue discussed at:
- https://github.com/ubuntu/yaru.dart/issues/251


Jupi007 (approved):
LGFM :+1: 

Thanks :)

Issue #549, opened 2023-01-19
Hero tags


jpnurmi:
I noticed that FloatingActionButton allows passing in a [hero tag](https://api.flutter.dev/flutter/material/FloatingActionButton/heroTag.html). We should do the same thing for YaruDetailPage & YaruTitleBar.

PR #553, opened 2023-01-20
YaruDetailPage: allow specifying the hero tag

jpnurmi:
No visual changes. A partial fix to #549. Something similar for YaruTitleBar comes separately.


Jupi007 (approved):
:+1:  Same as your second PR :)

Jupi007 (approved):
LGFM :+1: 

Simpler is better :D

jpnurmi:
Thanks for making me re-think what I was doing!

PR #554, opened 2023-01-20
YaruTitleBar: allow specifying the hero tag

jpnurmi:
The YaruTitleBar counterpart of #553.

Close: #549

### DIFF
--- a/lib/src/controls/yaru_title_bar.dart
+++ b/lib/src/controls/yaru_title_bar.dart
@@ -10,7 +10,7 @@ import 'yaru_title_bar_theme.dart';
 import 'yaru_window.dart';
 import 'yaru_window_control.dart';
 
-const _kYaruTitleBarHeroTag = '<_YaruTitleBar hero tag>';
+const _kYaruTitleBarHeroTag = '<YaruTitleBar hero tag>';
 
 /// A [Stack] of a [Widget] as [title] with a close button
 /// which pops the top-most route off the navigator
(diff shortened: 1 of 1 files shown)


Jupi007 (approved):
LGFM :+1: 

I suppose that the `toString` method of the object passed to `_YaruTitleBarHeroTag.id()` is always properly implemented.

jpnurmi:
In our case, it becomes "<_YaruDetailPageHeroTag YaruWindowTitleBar>" so it reads like "the detail page's hero tag for the app bar". We need the same tag for both title bars on the page transitioning in and out to achieve the effect that it stays in place. With custom hero tags you can either disable the whole thing if it's not suitable for your app structure, or you can adapt it to more complex scenarios e.g. only applying the effect when transitioning between certain pages.

Jupi007:
Okay, thanks for the explanation :)
It is fine so, can I merge?

jpnurmi:
Now that I think of it, I think the dynamic hero-tag formatting is pointless. We could just as well have a constant string?

Jupi007:
As far as I understood, a constant would be enough, indeed :)

jpnurmi:
Yeah, seems to work fine. What was I even thinking :man_facepalming:

jpnurmi:
how many attempts do i need to create a string constant

jpnurmi:
> how many attempts do i need to create a string constant

Aaaand I even managed to screw it up in the end???

[Screencast from 2023-01-20 19-54-19.webm](https://user-images.githubusercontent.com/140617/213782456-8c3ea8ab-2b89-4af9-8db5-3ef2eace3380.webm)

Issue #546, opened 2023-01-19
YaruBorderContainer wrong clipping


jpnurmi:
![image](https://user-images.githubusercontent.com/140617/213492406-6efc31e6-8089-41b1-ba35-884aa2b8ec88.png)

![image](https://user-images.githubusercontent.com/140617/213492657-002cf060-944d-43e1-bfda-a5ca5388bd72.png)

PR #226, opened 2022-09-25
Replace YaruColorPickerButton with YaruOptionButton.color()

jpnurmi:
For example, an option button with an arbitrary child widget:
```dart
YaruOptionButton(
  onPressed: () {},
  child: Icon(YaruIcons.search),
),
```
Or with a pre-made color disk:
```dart
YaruOptionButton.color(
  onPressed: () {},
  color: Theme.of(context).primaryColor,
),
```
image

PR #556, opened 2023-01-20
YaruBorderContainer: fix border & material clipping

jpnurmi:
| Before | After |
|---|---|
| image | image |

Fixes: #546


jpnurmi:
@Jupi007 thanks for the `BorderRadius` extension for calculating the inner radius :wink:

Jupi007:
Honestly I have mixed feeling about this 🤔

Looking at golden images, I think the previous border had a better contrast due to opacity stacking:

image

If you attentively look at the left side, it looks sharp, while on the right side, the similar colours makes the container edges blurry.

> @Jupi007 thanks for the BorderRadius extension for calculating the inner radius wink

Thank you ❤️

Feichtmeier:
We should avoid transparent borders. Sorry for introducing them just to get the correct shade of gray. The only places where they should stay are window, popup menu and dialog borders

jpnurmi:
The problem is that while the transparent border _might_ look ok'ish over something like a background highlight, the result is completely arbitrary because there could be anything else. It looks plain wrong when there's text :)

image

jpnurmi:
That is, even if the transparent border was correctly drawn on top. :D

Feichtmeier:
Let's just colour pick the colour and change in in yaru to be fully opaque

Jupi007:
To mitigate the problem I have shown upper, we should make the selected item background the same colour as the border.
So, either make the border brighter, or make the background darker, or something intermediate.

jpnurmi:
> Let's just colour pick the colour and change in in yaru to be fully opaque

Sounds good, performance-wise much better than double-clipping two different rounded corners :+1:

Jupi007:
> Let's just colour pick the colour and change in in yaru to be fully opaque

Yes, it should be opaque, in addition to the changes of this PR which move the border outside of the container (to avoid to lost 1px of content behind the border).
But if you look at the left side of my screenshot, the half-transparent border makes its colour be lighter on top of the selected item.
So there's exactly two colours. And as I said, the solution would be to have something unified between the border and the selected background.

Jupi007:
We need to do something similar to this: https://github.com/ubuntu/yaru.dart/pull/226

jpnurmi:
> To mitigate the problem I have shown upper, we should make the selected item background the same colour as the border. So, either make the border brighter, or make the background darker, or something intermediate.

The screenshot had hover highlight. Focus highlight, press, ripple, they all have slightly different shades.

Jupi007:
> The screenshot had hover highlight. Focus highlight, press, ripple, they all have slightly different shades.

Which screenshot do you mean? I picked this one [yaru_border_container-clip-dark.png](https://github.com/ubuntu/yaru_widgets.dart/pull/556/files#diff-1f0d340c9ab95173a08d1e09dc64a722b71bf1f2c4461326aaa855fde5c4086d)
Is this in hover state?

jpnurmi:
> > The screenshot had hover highlight. Focus highlight, press, ripple, they all have slightly different shades.
> 
> Which screenshot do you mean? I picked this one [yaru_border_container-clip-dark.png](https://github.com/ubuntu/yaru_widgets.dart/pull/556/files#diff-1f0d340c9ab95173a08d1e09dc64a722b71bf1f2c4461326aaa855fde5c4086d) Is this in hover state?

Hmm, I think that's actually pressed. :D The one in the PR description was hovered. Even though my point was that it's a generic container where users can put anything, including arbitrary scrollable content. :)

Jupi007:
> Hmm, I think that's actually pressed. :D The one in the PR description was hovered. Even though my point was that it's a generic container where users can put anything, including arbitrary scrollable content. :)

Okay, I get your point. I had forgotten that this was a generic container :D

Feichtmeier:
With the opaque divider now is this fixed?

PR #564, opened 2023-01-25
Update goldens with Flutter 3.7.0 & latest yaru

jpnurmi:


Feichtmeier (approved):
lgtm - interestingly some small text have a better contrast now, which was band aid fixed in software often

jpnurmi:
Looks like we should take care of yaru.dart first, because this one adds brown popup menus and white separators...

Jupi007:
Wtf, why does our own togglables are affected.
I remember that they only use the accent color 🤔

Issue #570, opened 2023-01-28
Navigation page pane layout delegate


Jupi007:
Implement a `YaruNavigationPagePaneLayoutDelegate`, which allows to control a `YaruNavigationPage` pane width.

See also: https://github.com/ubuntu-flutter-community/software/issues/895

PR #571, opened 2023-01-28
Allow YaruNavigationRailItem width control

Jupi007:
The current way of control a navigation page pane width weird because it is done by tiles themself.
For now I added a way to control this width, I'll think about something to control this size from the navigation page itself, without introduce breaking change later (in any case these changes are needed).

image

Related to #570

## Pull request checklist


Feichtmeier (approved):
ty!

Issue #583, opened 2023-02-05
YaruSection: unbounded height


jpnurmi:
If you don't pass any height, `YaruSection` eats up all it gets. Here's a _centered_ section:

![image](https://user-images.githubusercontent.com/140617/216808853-7e8e69d0-f68c-46bc-a4ec-c0de7f97ffcd.png)

```dart
import 'package:flutter/material.dart';
import 'package:yaru/yaru.dart';
import 'package:yaru_widgets/yaru_widgets.dart';

Future<void> main() async {
  await YaruWindowTitleBar.ensureInitialized();

  runApp(const MyApp());
}

class MyApp extends StatelessWidget {
  const MyApp({super.key});

  @override
  Widget build(BuildContext context) {
    return YaruTheme(
      builder: (context, yaru, child) {
        return MaterialApp(
          theme: yaru.theme,
          darkTheme: yaru.darkTheme,
          debugShowCheckedModeBanner: false,
          home: const Scaffold(
            appBar: YaruWindowTitleBar(),
            body: Center( // <==
              child: YaruSection(
                headline: Text('Section title'),
                child: Text('Section child'),
              ),
            ),
          ),
        );
      },
    );
  }
}
```

Feichtmeier:
that's not nice by yaru section! Incredibly greedy :laughing:

Feichtmeier:
it just needs a MainAxisSize parameter

Why is it inside "page" dir? And the pages are inside a different dir? :smile:

jpnurmi:
I also don't know why YaruBanner is in utils/, YaruTile in pages/, and YaruWindow in controls/... :man_shrugging: Maybe we can just flatten the tree?

Feichtmeier:
Good idea

PR #584, opened 2023-02-05
YaruSection: fix unbounded height

jpnurmi:
image

Close: #583

### DIFF
--- a/lib/src/pages/yaru_section.dart
+++ b/lib/src/pages/yaru_section.dart
@@ -45,6 +45,7 @@ class YaruSection extends StatelessWidget {
       padding: padding,
       margin: margin,
       child: Column(
+        mainAxisSize: MainAxisSize.min,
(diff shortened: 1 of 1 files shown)


lib/src/pages/yaru_section.dart
```diff
@@ -45,6 +45,7 @@ class YaruSection extends StatelessWidget {
       padding: padding,
       margin: margin,
       child: Column(
+        mainAxisSize: MainAxisSize.min,
```
Feichtmeier, line 48: Could you make it a parameter with this as default? Maybe there are cases where we want it to expand?

Feichtmeier (changes requested):

lib/src/pages/yaru_section.dart
jpnurmi: you can expand the content as you wish. it's intentionally wrapped with flexible

PR #593, opened 2023-02-05
Introduce YaruPageIndicator

### LINKED ISSUE #582: Page indicator
Could we have this as a standalone widget? It would be useful for non-carousel page views and wizards etc. 

image

Jupi007:
[Capture vidéo du 2023-02-05 19-43-56.webm](https://user-images.githubusercontent.com/36476595/216838646-6e308dd3-025f-45b7-bc08-4259bd5f509c.webm)

Fixes #582

## Pull request checklist

### DIFF
--- a/lib/src/widgets/yaru_page_indicator.dart
+++ b/lib/src/widgets/yaru_page_indicator.dart
@@ -1,0 +1,111 @@
+import 'package:flutter/material.dart';
+
+/// A Yaru page indicator.
+///
+/// If there's enough space, it will be rendered into a line of dots,
+/// if not, it will be rendered into a text-based indicator.
+class YaruPageIndicator extends StatelessWidget {
+  /// Create a [YaruPageIndicator].
+  const YaruPageIndicator({
+    super.key,
+    required this.length,
+    required this.page,
+    this.animationDuration = Duration.zero,
+    this.animationCurve = Curves.linear,
+    this.onDotTap,
+  }) : assert(page >= 0 && page <= length - 1);
+
+  /// Determine the number of pages.
+  final int length;
+
+  /// Current page index.
+  /// This value should be clamped between 0 and [length] - 1
+  final int page;
+
+  /// Duration of a transition between two dots.
+  /// Use [Duration.zero] (defaults) to disable transition.
+  final Duration animationDuration;
+
+  /// Curve used in a transition between two dots.
+  final Curve animationCurve;
+
+  /// Callback called when tapping a dot.
+  /// It passes the tapped page index as parameter.
+  final Function(int page)? onDotTap;
+
+  @override
+  Widget build(BuildContext context) {
+    return LayoutBuilder(
+      builder: (context, constraints) {
+        const dotSize = 12.0;
+
+        for (final layout in [
+          [48.0, constraints.maxWidth / 2],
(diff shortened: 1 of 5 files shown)


lib/src/widgets/yaru_page_indicator.dart
```diff
@@ -0,0 +1,111 @@
...
+  /// Duration of a transition between two dots.
+  /// Use [Duration.zero] (defaults) to disable transition.
+  final Duration animationDuration;
+
+  /// Curve used in a transition between two dots.
+  final Curve animationCurve;
+
+  /// Callback called when tapping a dot.
+  /// It passes the tapped page index as parameter.
+  final Function(int page)? onDotTap;
```
jpnurmi, line 34: could it be called "onTap"? "dot" is an unnecessary description of the style we have right now :)

```suggestion
  final ValueChanged<int>? onTap;
```
- https://api.flutter.dev/flutter/material/TabBar/onTap.html
- https://api.flutter.dev/flutter/material/BottomNavigationBar/onTap.html

lib/src/widgets/yaru_page_indicator.dart
```diff
@@ -0,0 +1,111 @@
...
+    this.animationCurve = Curves.linear,
+    this.onDotTap,
+  }) : assert(page >= 0 && page <= length - 1);
+
+  /// Determine the number of pages.
+  final int length;
+
+  /// Current page index.
+  /// This value should be clamped between 0 and [length] - 1
+  final int page;
```
jpnurmi, line 23: could it be called "index" to follow the same naming as YaruPageController?

- https://pub.dev/documentation/yaru_widgets/latest/yaru_widgets/YaruPageController/index.html

jpnurmi (commented):
i left a couple of questions about the names but otherwise looks good!

lib/src/widgets/yaru_page_indicator.dart
Jupi007: Well I disagree, because this indicator is a row of dot, and it can be also just a `Text()`.
`onDotTap` makes it obvious that it is a callback only for a single dot.
Imo `onTap` means this is a callback for the full indicator onTap event :)

Jupi007: Sorry, I just saw your links 😅
You're right so!

lib/src/widgets/yaru_page_indicator.dart
Jupi007: Well, I used `page` because `PageController`, also`YaruCarouselController` who extends it, and `YaruCarousel` are using this name everywhere.

If I rename it here, should I rename it everywhere?

jpnurmi: No strong opinion on this :)

jpnurmi (approved):
LGTM :+1:

Jupi007:
Thanks for the review @jpnurmi :)

jpnurmi:
Thanks for extracting it!

This will come in handy in the upcoming master-detail example, and we might even use this in the installer if we just can figure out a reasonable way to calculate the wizard steps. :)

Jupi007:
Cool, I can't wait to see it in action :D

PR #616, opened 2023-02-12
YaruPopupMenuButton: make radius changeable

Feichtmeier:
## Pull request checklist

- [X] This PR does not introduce visual changes, **or**
  - I ran `flutter test --update-goldens` and committed the changes if there were any, **or**
  - I added before/after/light/dark screenshots if the visual changes I made were not covered by golden tests.
    | |Before|After|
    |-|-|-|
    |Light| | |
    |Dark| | |

PR #620, opened 2023-02-14
YaruPopupMenuButton: replace borderRadius with style

jpnurmi:
For example:
```
YaruPopupMenuButton<MyEnum>(
  style: OutlinedButton.styleFrom(
    shape: RoundedRectangleBorder(
      side: const BorderSide(color: Colors.red, width: 2),
      borderRadius: BorderRadius.circular(20),
    ),
  ),
  // ...
),
```

image

Ref: #616


Feichtmeier (approved):
yes! ty

PR #624, opened 2023-02-22
YaruPageIndicator: add custom scale parameters

### LINKED ISSUE #622: YaruPageIndicator: allow customization?
The page indicator currently has hard-coded dot size and spacing. It would be nice to have some control over this (maybe just a scaling factor for each of those?).

See discussion here: https://github.com/canonical/ubuntu-desktop-installer/pull/1445

d-loose:
Allows to customize the size and spacing of `YaruPageIndicator`'s dots:

[Screencast from 2023-02-22 12-48-53.webm](https://user-images.githubusercontent.com/113362648/220611821-6b85ab12-973f-463e-bf70-383df1521b2e.webm)

Fix #622

## Pull request checklist

    | |Before|After|
    |-|-|-|
    |Light| | |
    |Dark| | |

### DIFF
--- a/example/lib/pages/page_indicator.dart
+++ b/example/lib/pages/page_indicator.dart
@@ -12,8 +12,8 @@ class PageIndicatorPage extends StatefulWidget {
 class _PageIndicatorPageState extends State<PageIndicatorPage> {
   int _page = 0;
   int _length = 5;
-  double _dotSizeFactor = 1.0;
-  double _dotSpacingFactor = 1.0;
+  double _dotSize = 12.0;
+  double _dotSpacing = 48.0;
 
   @override
   Widget build(BuildContext context) {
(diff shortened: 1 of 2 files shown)


Feichtmeier:
Looks good because it works :smile: 

As I do not care enough for the implementation detail let @jpnurmi and @Jupi007 double check as they are usually a bit more nit-picky

Jupi007 (changes requested):
I agree that provide a way to customize dot size and spacing is a great idea!

But the way I implemented the current spacing was preventing layout overflow, while this PR doesn't provide any kind of security:

https://user-images.githubusercontent.com/36476595/220619199-d985504b-7fb6-45dd-8443-6e509300a628.mp4

@d-loose Can you implement a way to avoid overflow?

Also it would be nice that your new properties could not be factors, but real size:

```dart
final double dotSize = 12.0;
final double dotSpacingTarget; // target, because it can be stretched cause to the overflow guard
```

It is more intuitive in that way, and so we don't hide those parameters with some obscure internal stuff.

jpnurmi:
As a user, I'd rather specify the desired size and spacing than a factor because I don't know what the factor applies to :)

d-loose:
Thanks!
Vertical constraints are respected again, and I've changed the parameters to accept absolute values.

[Screencast from 2023-02-22 13-51-39.webm](https://user-images.githubusercontent.com/113362648/220624980-4345fb7b-c8a7-4cf3-97e8-3d11503bf82a.webm)

Jupi007 (approved):
I can't actually test it, but if you ensure me there's no more vertical overflow, then LGFM 🙂 👍

d-loose:
I can :)
It only happened before, because I recklessly rescaled `maxWidth` :laughing:

PR #632, opened 2023-02-24
Introduce YaruPageIndicatorTheme

### LINKED ISSUE #631: YaruPageIndicatorTheme
We would like to customize `YaruPageIndicator` to look like this in the [installer](https://github.com/canonical/ubuntu-desktop-installer/pull/1445#issuecomment-1441899274):

image

Perhaps we could expose some `BoxDecoration` properties such as shape, color, and border for the different states (current vs. past vs. future) via `YaruPageIndicatorTheme(Data)`?

Jupi007:
I used a "decoration builder" because it allows to do special things like this:

https://user-images.githubusercontent.com/36476595/221298238-c784a71b-168b-4419-a654-15f4045629d1.mp4

Fixes #631

## Pull request checklist

### DIFF
--- a/lib/src/widgets/yaru_page_indicator.dart
+++ b/lib/src/widgets/yaru_page_indicator.dart
@@ -3,7 +3,7 @@ import 'package:flutter/material.dart';
 import 'yaru_carousel.dart';
 import 'yaru_page_indicator_theme.dart';
 
-typedef DotDecorationBuilder = Decoration Function(
+typedef YaruDotDecorationBuilder = Decoration Function(
   int index,
 );
 
@@ -55,7 +55,7 @@ class YaruPageIndicator extends StatelessWidget {
   final double dotSpacing;
 
   /// Decoration of the dots.
-  final DotDecorationBuilder? dotDecorationBuilder;
+  final YaruDotDecorationBuilder? dotDecorationBuilder;
 
   @override
   Widget build(BuildContext context) {
--- a/lib/src/widgets/yaru_page_indicator.dart
+++ b/lib/src/widgets/yaru_page_indicator.dart
@@ -5,6 +5,8 @@ import 'yaru_page_indicator_theme.dart';
 
 typedef YaruDotDecorationBuilder = Decoration Function(
   int index,
+  int selectedIndex,
+  int length,
 );
 
 /// A responsive page indicator.
@@ -108,7 +110,7 @@ class YaruPageIndicator extends StatelessWidget {
       mainAxisSize: MainAxisSize.min,
       children: List<Widget>.generate(length, (index) {
         final dotDecoration = dotDecorationBuilder != null
-            ? dotDecorationBuilder.call(index)
+            ? dotDecorationBuilder.call(index, page, length)
             : BoxDecoration(
                 color: page == index
                     ? theme.colorScheme.primary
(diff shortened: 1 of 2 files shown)


jpnurmi (commented):
very cool! we might need to somehow communicate the current index to make it possible compare the "distance" to the current one?

lib/src/widgets/yaru_page_indicator.dart
```diff
@@ -1,6 +1,11 @@
 import 'package:flutter/material.dart';
 
 import 'yaru_carousel.dart';
+import 'yaru_page_indicator_theme.dart';
+
+typedef DotDecorationBuilder = Decoration Function(
+  int index,
```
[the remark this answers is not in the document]
Jupi007: I was asking this question to myself.
But yeah we definitively should, because the theme definition can be far away from the page indicator itself, and so we can't know the current index and the page length.

lib/src/widgets/yaru_page_indicator.dart
```diff
@@ -1,6 +1,11 @@
 import 'package:flutter/material.dart';
 
 import 'yaru_carousel.dart';
+import 'yaru_page_indicator_theme.dart';
+
+typedef DotDecorationBuilder = Decoration Function(
```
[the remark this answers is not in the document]
Jupi007: Oh, yes your right :+1:

lib/src/widgets/yaru_page_indicator.dart
```diff
@@ -48,8 +56,17 @@ class YaruPageIndicator extends StatelessWidget {
...
+  /// Decoration of the dots.
+  final YaruDotDecorationBuilder? dotDecorationBuilder;
+
   @override
   Widget build(BuildContext context) {
+    final theme = Theme.of(context);
+    final indicatorTheme = YaruPageIndicatorTheme.of(context);
+
+    final dotSize = indicatorTheme?.dotSize ?? this.dotSize;
+    final dotSpacing = indicatorTheme?.dotSpacing ?? this.dotSpacing;
```
jpnurmi, line 68: we need to swap these around. a local widget property should win over a theme-level property. perhaps we need to make the widget properties nullable?

lib/src/widgets/yaru_page_indicator.dart
```diff
@@ -61,30 +78,45 @@ class YaruPageIndicator extends StatelessWidget {
...
+    YaruPageIndicatorThemeData? indicatorTheme,
     double dotSize,
     double dotSpacing,
   ) {
+    final dotDecorationBuilder =
+        indicatorTheme?.dotDecorationBuilder ?? this.dotDecorationBuilder;
+    final animationDuration =
+        indicatorTheme?.animationDuration ?? this.animationDuration;
+    final animationCurve =
+        indicatorTheme?.animationCurve ?? this.animationCurve;
```
jpnurmi, line 106: same here, `this.foo` needs to come before `theme.foo`

lib/src/widgets/yaru_page_indicator.dart
Jupi007: Oh, right.
It's a bit of a stupid mistake from myself to invert that 😅

I'll also make the widget properties nullable :+1:

jpnurmi (approved):
The builder was a nice idea! LGTM :+1:

Jupi007:
Thanks 😊

PR #635, opened 2023-02-25
Example: add macOS

jpnurmi:
Generated with `flutter create --platforms macos` and the same window size as we have on Linux was applied in the .xib file.

image


Jupi007:
@jpnurmi I haven't got any apple device at home, so I don't think I'm the good person to review this 😅

Maybe @Feichtmeier or @madsrh I think they got some mac book.

jpnurmi:
P.S. It's nice to be able to run the example on macOS but my main motivation is a macOS test build in the CI. :)

Jupi007:
Oh okay.
I take a look at the changes, and from my point of view it looks good :+1: 

The amount of file required to have a macOS app is impressive!

PR #636, opened 2023-02-25
Example: add Windows

jpnurmi:
This allows us to a) easily test the window title bar on Windows and b) add a Windows build test into to CI pipeline.

Note: The files were automatically generated by `flutter create --platforms windows`. The only manual change was to set the default window size in `main.cpp` to match other platforms:
```cpp
Win32Window::Size size(700, 720);
```

yaru_widgets


Jupi007 (approved):
LGTM 👍

PR #294, opened 2022-10-12
Golden YaruIconButton test

jpnurmi:
This test catches unintentional visual changes. Whenever pushing intentional visual changes, one would have to update the golden images by running:
```
flutter test --update-goldens
```
P.S. GitHub shows the before and after images side by side in the PR diff. :wink:

PR #637, opened 2023-02-25
Example: upgrade handy_window to ^0.3.0

jpnurmi:
Large shadows were noticed in https://github.com/ubuntu/yaru.dart/issues/294#issuecomment-1445031772.


Jupi007 (approved):
LGFM 👍

PR #642, opened 2023-02-26
Add `YaruMasterDetailPage.orientationOf()` and `appBarBuilder`

jpnurmi:

### DIFF
--- a/lib/src/widgets/master_detail/yaru_master_detail_page.dart
+++ b/lib/src/widgets/master_detail/yaru_master_detail_page.dart
@@ -114,9 +114,15 @@ class YaruMasterDetailPage extends StatefulWidget {
   /// Returns the orientation of the [YaruMasterDetailPage] that most tightly
   /// encloses the given context.
   static Orientation orientationOf(BuildContext context) {
-    return context
-        .dependOnInheritedWidgetOfExactType<_YaruMasterDetailLayoutScope>()!
-        .orientation;
+    return maybeOrientationOf(context)!;
+  }
+
+  /// Returns the orientation of the [YaruMasterDetailPage] that most tightly
+  /// encloses the given context.
+  static Orientation? maybeOrientationOf(BuildContext context) {
+    final scope = context
+        .dependOnInheritedWidgetOfExactType<_YaruMasterDetailLayoutScope>();
+    return scope?.orientation;
   }
 
   @override
--- a/lib/src/widgets/master_detail/yaru_master_detail_page.dart
+++ b/lib/src/widgets/master_detail/yaru_master_detail_page.dart
@@ -165,31 +165,25 @@ class _YaruMasterDetailPageState extends State<YaruMasterDetailPage> {
   Widget build(BuildContext context) {
     return Scaffold(
       body: _YaruMasterDetailLayoutBuilder(
-        builder: (context, orientation) {
-          switch (orientation) {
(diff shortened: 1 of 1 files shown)


Jupi007 (approved):
Very cool, LGFM :+1: 

_you just need to fix a missing comma_

jpnurmi:
@Feichtmeier does this help with the back button issue? i have a feeling it doesn't because if `Navigator.of(context).canPop()` didn't work for you, it sounds like you're building the title bar outside the master-detail page so you can't access the orientation either via an inherited widget like this either?

Feichtmeier:
ah you mean because

```dart
return Column(
      children: [
        appBar,
        Expanded(
          child: YaruMasterDetailPage(
          ```
          
          ?
          
          Well i don't know how to get this layout otherwise. Should I wrap it in a scaffold?

**Comment by jpnurmi** [comment]
I'm still trying to understand the requirements. :)

The original question was how to query MD orientation and this PR adds that mechanism but I'm afraid it's not possible to access an inherited widget from where you'd like to, above the master detail page.

We could make it possible to query the current orientation from the controller, though.

```dart
class YaruMasterDetailController extends YaruPageController {
  Orientation get orientation => ...
}
```

Feichtmeier:
> I'm still trying to understand the requirements. :)
> 
> The original question was how to query MD orientation and this PR adds that mechanism but I'm afraid it's not possible to access an inherited widget from where you'd like to, above the master detail page.
> 
> We could make it possible to query the current orientation from the controller, though.
> 
> ```dart
> class YaruMasterDetailController extends YaruPageController {
>   Orientation get orientation => ...
> }
> ```

Sounds enough for my case. But is this layout really so weird? 😄 Alternatively I could maybe wrap it in a scaffold?

jpnurmi:
It's not the layout being weird or you doing anything wrong. We definitely want to support the layout you have. It's just impossible in Flutter to access an _inherited_ widget from the context above so we need to make it possible to get the information in some other way.

```dart
Scaffold(
  appBar: YaruWindowTitleBar(
    // context OUTSIDE YaruMasterDetailPage
    // - Navigator.of(context) finds the ROOT navigator, not the MD navigator
    // - YaruMasterDetailPage.of(context) doesn't find anything
  ),
  body: YaruMasterDetailPage(
    pageBuilder: (context, index) {
      // context INSIDE YaruMasterDetailPage
      // - Navigator.of(context) finds the MD navigator
      // - YaruMasterDetailPage.of(context) is OK
    },
  ),
)
```

Feichtmeier:
ah right! 
I test this PR now

Feichtmeier:
> @Feichtmeier does this help with the back button issue? i have a feeling it doesn't because if `Navigator.of(context).canPop()` didn't work for you, it sounds like you're building the title bar outside the master-detail page so you can't query the MD orientation via an inherited widget like this either?

confirming. as the element is not reachable from down there 
Thanks for trying though! If this is still an improvement you can ofc still merge as you like

jpnurmi:
@Jupi007 @Feichtmeier do we have any use case for this?

Jupi007:
Imo, this can be useful in some situations.
Ex: I you want to make something different in the detail view depending on the layout.

jpnurmi:
Thanks for chiming in. I started feeling somehow unsure because this didn't solve the original issue. :)

PR #643, opened 2023-02-26
`YaruPortraitLayout`: add missing hero controller

jpnurmi:
It should also have one, just like `YaruLandscapeLayout` and `YaruNavigationPage` do.


Jupi007:
I suppose this is a patch?
Because so you need to target the `release` branch 😉

Jupi007 (approved):
LGTM 👍

Issue #85, opened 2022-02-11
Remove hardcoded search behaviour


Feichtmeier:
Let users of this package decide how the search works.
This probably needs a reworked search bar for the master details page

PR #648, opened 2023-03-04
Check/Radio/Switch/Button/ListTile: allow customizing mouse cursor

jpnurmi:
- https://github.com/canonical/ubuntu-desktop-installer/issues/1532

Ref: ubuntu/yaru.dart#85

### DIFF
--- a/lib/src/widgets/yaru_check_button.dart
+++ b/lib/src/widgets/yaru_check_button.dart
@@ -52,10 +52,7 @@ class YaruCheckButton extends StatelessWidget {
     final mouseCursor = this.mouseCursor ??
         YaruToggleButtonTheme.of(context)
             ?.mouseCursor
-            ?.resolve({if (onChanged == null) MaterialState.disabled}) ??
-        (onChanged != null
-            ? SystemMouseCursors.click
-            : SystemMouseCursors.basic);
+            ?.resolve({if (onChanged == null) MaterialState.disabled});
 
     return YaruToggleButton(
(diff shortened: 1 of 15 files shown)


Jupi007 (changes requested):
This seems to looks good :)
But can you also add the `mouseCursor` property to checkbox/radio/switch/togglable theme classes?

**Edit:** I suppose you also plan a PR to change the default cursor?

jpnurmi:
> But can you also add the `mouseCursor` property to checkbox/radio/switch/togglable theme classes?

Thanks, done!

> **Edit:** I suppose you also plan a PR to change the default cursor?

It's a different discussion but let's start by making them customizable from the outside, at least. :)

I used to feel pretty strong against the hand cursors which is why I spent so much time contributing changes to Flutter to make it themable. The problem is that there are a) many legacy widgets that are missing themes altogether or cannot be made themable without breaking things and b) there are new widgets coming up in every release that are missing themable mouse cursors. As a result, you may end up with a strange mix where some widgets have hand and some basic cursors, and you might have to finish it off by applying some hacks on the app side. Therefore it might be hard to make it the default until Flutter offers a clean way to change the clickable cursor everywhere.

Jupi007 (approved):
LGTM 👍

Jupi007:
I understand.
But to be very honest, now I pretty much like the usage of the pointer cursor in clickable area 😅
It adds an additional semantic information.

PR #658, opened 2023-03-05
Resolve mouse cursor widget properties

jpnurmi:
I was looking at some old Flutter PRs for themable mouse cursors and noticed that the `MouseCursor` type of widget properties were not used "as is" but also resolved against the widget state with help of `MaterialStateProperty.resolveAs`. The Flutter team recommended doing it this way to support `MaterialStateMouseCursor` in place of `MouseCursor`.


Jupi007 (approved):
Fine LGTM 👍

Issue #659, opened 2023-03-05
Check/Radio/SwitchButton hover highlight


jpnurmi:
Perhaps we could highlight the Check/Radio/Switch indicators when hovering over the Check/Radio/SwitchButton label as illustrated by the Materia GTK3 theme below.

[Screencast from 2023-03-05 22-34-00.webm](https://user-images.githubusercontent.com/140617/222987312-56c29d6d-dc66-494a-bfae-2cdac47798e8.webm)

PR #661, opened 2023-03-06
Make YaruToggleButton subclasses stateful

jpnurmi:
This is a noisy preparation step for #659 to make it possible for `YaruToggleButton` subclasses to create and dispose an internal `MaterialStatesController`.


Jupi007 (approved):
LGFM 👍

Issue #305, opened 2022-10-17
Desktop Web: separator of MD page is super white in the dark theme


Feichtmeier:
Combination: dark theme + desktop + web (FF and chrome)

![grafik](https://user-images.githubusercontent.com/15329494/196217166-5f76023b-32fb-4547-b19b-ff9309f1219b.png)

Does not happen in the compact layout 🤷

Jupi007:
Very weird, I also noticed that the separator color is changing depending on the window focus (on desktop).

Feichtmeier:
Maybe it's related to the drawer thing 🙂
I can check maybe this evening

PR #670, opened 2023-03-11
YaruWindowControl: add subtle outline for high-contrast themes

jpnurmi:
The outline is not really noticeable except with high-contrast themes to make the window controls stand out for accessibility reasons.

|   | Before | After |
|---|--------|-------|
| Light HC (experimental) | image | image |
| Dark HC (experimental) | image | image |
| Light | image | image |
| Dark | image | image |

Ref: ubuntu/yaru.dart#305


jpnurmi:
GCC for reference:

| Light | Dark |
|---|---|
| image | image |

Feichtmeier:
Holy moly, 😂
You had me a little scared with "not really noticeable". But they are not noticeable as far as I can see which is the only possible way to merge this for normal themes. Can anyone see anything?!?

jpnurmi:
I was like "he's gonna murder me" if I even propose an outline so I was relieved I couldn't see it with my bare eyes with the normal themes. :D

jpnurmi:
Actually, one can notice it when disabled. The screenshot is not ideal though because it's taken against a white background.

image

jpnurmi:
Here's with minimize & maximize disabled:

| Light | Dark |
|---|---|
| image | image |

Feichtmeier:
Hm can you make this improvement on high contrast without affecting the other themes? Our gnome window controls also have no outline, regarding of enabled/disabled

Jupi007:
I agree, there's shouldn't be any outline in default theme.

But it is a great idea for the HC mode 👍

PR #672, opened 2023-03-11
YaruPageIndicator: restore compatibility

jpnurmi:
Restore the old `dotSize` & `dotSpacing` arguments for the default constructor, and introduce an alternative named constructor for the new item size builder and layout delegate. This would be enough to avoid breaking the installer that is already using `dotSize` & `dotSpacing`. WDYT?

### DIFF
--- a/example/lib/pages/page_indicator.dart
+++ b/example/lib/pages/page_indicator.dart
@@ -20,7 +20,7 @@ class _PageIndicatorPageState extends State<PageIndicatorPage> {
     return ListView(
       padding: const EdgeInsets.all(kYaruPagePadding),
       children: [
-        YaruPageIndicator.delegate(
+        YaruPageIndicator.builder(
           length: _length,
           page: _page,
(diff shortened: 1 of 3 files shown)


Jupi007 (approved):
Great idea!
I didn't thought about mitigate the backward compatibility.

jpnurmi:
What about the name? Do you like `YaruPageIndicator.builder()` or `YaruPageIndicator.delegate()` more? Perhaps the better is actually better because the default constructor still takes one builder i.e. `itemBuilder`... :thinking:

Jupi007:
Yeah, it makes sense to name it `YaruPageIndicator.delegate()` :+1:

jpnurmi:
> **Edit:** or drop the builder parameters from the default constructor, and leave their access in `YaruPageIndicator.builder()`.

Great idea! :+1: I forgot it was also a new argument. Now the builder naming makes sense. :)

Jupi007 (approved):
Great, LGTM :+1:

PR #673, opened 2023-03-13
YaruAutocomplete: calculate matching options width by default

jpnurmi:
Flutter's [Autocomplete](https://api.flutter.dev/flutter/material/Autocomplete-class.html) widget expands the options popup to all available width. Given that it looks somehow unusual on desktop, we made it possible to override the width in https://github.com/ubuntu/yaru_widgets.dart/pull/668.

Most often, the autocomplete widget can be simply wrapped with a LayoutBuilder to get the width _but_ things get more complicated when trying to use it in a dialog because LayoutBuilder naturally doesn't support intrinsic sizes. Therefore, I'm proposing to make this the default behavior and let `YaruAutocomplete` calculate the matching width under the hood. The old behavior is still possible by e.g. passing the screen width or simply `optionsWidth: double.infinity`.

| Before | After |
|---|---|
| Screenshot from 2023-03-13 10-44-25 | Screenshot from 2023-03-13 10-44-08 |

### DIFF
--- a/lib/src/widgets/yaru_autocomplete.dart
+++ b/lib/src/widgets/yaru_autocomplete.dart
@@ -3,7 +3,7 @@ import 'package:flutter/scheduler.dart';
 
 import 'yaru_border_container.dart';
 
-class YaruAutocomplete<T extends Object> extends StatelessWidget {
+class YaruAutocomplete<T extends Object> extends StatefulWidget {
   const YaruAutocomplete({
     super.key,
     required this.optionsBuilder,
@@ -32,7 +32,11 @@ class YaruAutocomplete<T extends Object> extends StatelessWidget {
 
   /// The width used for the default options list widget.
   ///
-  /// The default value is null.
+  /// The default value is null which means the width will be the same as the
+  /// width of the [YaruAutocomplete] widget.
+  ///
+  /// Note: Passing `double.infinity` makes the options expand to the width of
+  /// the screen similarly to the Material design Autocomplete widget.
   final double? optionsWidth;
 
   /// The maximum height used for the options list widget.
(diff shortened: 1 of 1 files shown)


Jupi007 (commented):
I see you're using a post frame callback, but isn't it creating a size jump during a frame?

Also, there isn't any page for this widget in the example app, is it intentional?

jpnurmi:
> I see you're using a post frame callback, but isn't it creating a size jump during a frame?

There's no render box until the widget has been rendered but notice that the popup is not visible during initState() anyway. The same approach is used by some Flutter widgets that deal with popups. :)

> Also, there isn't any page for this widget in the example app, is it intentional?

I dropped it for the time being because it was too ugly to be the starting page :grinning: 

https://github.com/ubuntu/yaru_widgets.dart/pull/668

Jupi007 (approved):
> There's no render box until the widget has been rendered but notice that the popup is not visible during initState() anyway. The same approach is used by some Flutter widgets that deal with popups. :)

Okay, that's perfect so 👍 🙂

> I dropped it for the time being because it was too ugly to be the starting page grinning

Oh, I didn't noticed this commit, maybe would it be a better solution to make the default page at a different index, is it possible currently?

jpnurmi:
For reference, LayoutBuilder works fine here:

image

But it explodes when used in a dialog like this that relies on intrinsic sizes:

image

Jupi007:
> But it explodes when used in a dialog like this that relies on intrinsic sizes:

Wow, yeah this is not how it should look 😅

Jupi007:
Imo the banner page is the best candidate for a default page :+1: 
We can set the default index with a `YaruPageController`.

PR #674, opened 2023-03-13
Example: add autocomplete

jpnurmi:
NOTE: The banner page is the default because the autocomplete page is not the most attractive page... :P

image

Screenshot from 2023-03-13 11-26-41

### DIFF
--- a/example/lib/example.dart
+++ b/example/lib/example.dart
@@ -54,6 +54,7 @@ class _MasterDetailPage extends StatelessWidget {
   @override
   Widget build(BuildContext context) {
     return YaruMasterDetailPage(
+      initialIndex: 1,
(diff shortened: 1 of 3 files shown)


jpnurmi:
I set `initialIndex: 1` because we have a different set of pages in MD and navi-rail modes so one cannot share the controller due to different length... oops

Jupi007:
> @Jupi007 I set initialIndex: 1 as a quick solution because we have a different set of pages in MD and navi-rail modes so one cannot share the controller due to different length... oops

Indeed, I am also experiencing this problem :D
The initial index seems fine to me :+1:

Jupi007 (approved):
LGTM, thanks :)

PR #676, opened 2023-03-14
Add `emptyBuilder`

### LINKED ISSUE #669: YaruMasterDetailPage: "empty" builder?
It could be a convenient way to handle empty state if the list of pages is asynchronously constructed, for example. Throwing an exception forces conditional building which doesn't look as nice...

```dart
body: pages.isEmpty
  ? YaruCircularProgressIndicator()
  : YaruMasterDetailPage(
      length: pages.length,
      pageBuilder: ...
    ),
```
vs.
```dart
body: YaruMasterDetailPage(
  length: pages.length,
  pageBuilder: ...
  emptyBuilder: (context) => YaruCircularProgressIndicator(),
),
```

See also:
- https://github.com/canonical/firmware-updater/issues/161

d-loose:
Close #669

### DIFF
--- a/lib/src/widgets/master_detail/yaru_master_detail_page.dart
+++ b/lib/src/widgets/master_detail/yaru_master_detail_page.dart
@@ -167,10 +167,10 @@ class _YaruMasterDetailPageState extends State<YaruMasterDetailPage> {
 
   @override
   Widget build(BuildContext context) {
-    return widget.length == 0 || widget.controller?.length == 0
-        ? widget.emptyBuilder?.call(context) ?? const SizedBox.shrink()
-        : Scaffold(
-            body: _YaruMasterDetailLayoutBuilder(
+    return Scaffold(
+      body: widget.length == 0 || widget.controller?.length == 0
+          ? widget.emptyBuilder?.call(context) ?? const SizedBox.shrink()
+          : _YaruMasterDetailLayoutBuilder(
               portrait: (context) => YaruPortraitLayout(
                 tileBuilder: widget.tileBuilder,
                 pageBuilder: widget.pageBuilder,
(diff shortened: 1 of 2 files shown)


lib/src/widgets/master_detail/yaru_master_detail_page.dart
```diff
@@ -163,29 +167,31 @@ class _YaruMasterDetailPageState extends State<YaruMasterDetailPage> {
...
-          previousPaneWidth: _previousPaneWidth,
-          onLeftPaneWidthChange: (width) => _previousPaneWidth = width,
-          appBar: widget.appBar ?? widget.appBarBuilder?.call(context),
-          bottomBar: widget.bottomBar,
-          controller: _controller,
-        ),
-      ),
-    );
+    return widget.length == 0 || widget.controller?.length == 0
+        ? widget.emptyBuilder?.call(context) ?? const SizedBox.shrink()
```
jpnurmi, line 171: what do you think about moving this inside the scaffold?

i'm thinking about the kind of structure where you have `YaruMasterDetailPage` right below `MaterialApp`. if we have the empty builder inside the `Scaffold`, we wouldn't get errors about missing `Material` etc.

d-loose: Good point. How can we handle this in the `YaruNavigationPage`? We have `LayoutBuilder`->`SafeArea`->`Scaffold` there. Is the `SafeArea` needed for anything, and can we just move the `Scaffold` to the top? (On first glance it seems to work, but maybe there are reasons for this widget hierarchy I'm not aware of)

jpnurmi: Not sure where SafeArea came from and even less sure if it does anything on desktop platforms...

jpnurmi (approved):
LGTM :+1: but let's give @Feichtmeier and @Jupi007 a chance to chime in

Jupi007 (approved):
Great, LGTM 👍

Feichtmeier (approved):

PR #685, opened 2023-03-27
YaruWindowTitleBar.ensureInitialized(): clear cached window states

jpnurmi:
This allows removing any previously cached window state during the test setup phase.


d-loose (approved):
:+1:

PR #692, opened 2023-04-13
Expansion_panel

Feichtmeier:
Combines the magic of yaru expandable and some lists to an expansion panel that collapses the other children if one is expanded. Added some parameters of yarubordercontainer and expandable.

Do you think this is enough or should I add more parameters @Jupi007 ?

https://user-images.githubusercontent.com/15329494/231852738-da272acf-89f5-4e37-a7be-420cfffe3067.mov


Jupi007 (approved):
I don't see anything missing, looks very great!
LGFM 🚀

PR #701, opened 2023-05-29
chore: update yaru deps, dart and flutter

Feichtmeier:
- update to dart 3 and flutter 3.10
- update to yaru 0.8.0, yaru icons 1.0.4 and yaru window 0.1.2
- change binding.window.* to view.* in golden tests

PR #702, opened 2023-05-30
chore: migrate to Flutter 3.10 and Dart 3.0

jpnurmi:


jpnurmi:
I have no idea what was the problem in #701 because this looks very similar...

Feichtmeier (approved):
Thank you! I wonder what I did wrong in my attempt

PR #712, opened 2023-06-18
Drop double `YaruMasterTileScope` retrieval

### LINKED ISSUE #711: `YaruMasterTileScope` is retrieved twice
`YaruMasterTileScope` is retrieved a first time at the beginning of the `build` method:

https://github.com/ubuntu/yaru_widgets.dart/blob/735079409528cb16f56a817c4e004cb00fbc1654/lib/src/widgets/master_detail/yaru_master_tile.dart#L44

And then a second time inside the `onTap` callback, from the exact same `context` as the first one:

https://github.com/ubuntu/yaru_widgets.dart/blob/735079409528cb16f56a817c4e004cb00fbc1654/lib/src/widgets/master_detail/yaru_master_tile.dart#L78-L82

Jupi007:
Fixes #711

### DIFF
--- a/lib/src/widgets/master_detail/yaru_master_tile.dart
+++ b/lib/src/widgets/master_detail/yaru_master_tile.dart
@@ -76,7 +76,6 @@ class YaruMasterTile extends StatelessWidget {
           trailing: trailing,
           selected: isSelected,
           onTap: () {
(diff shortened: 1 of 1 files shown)


jpnurmi (approved):
LGTM :+1:

PR #714, opened 2023-06-22
feat: add a stack option to YaruChoiceChipBar

Feichtmeier:
This introduces a third option "stack" for the choice chip bar. 

https://github.com/ubuntu/yaru_widgets.dart/assets/15329494/03a7bc80-f5cb-410e-98ed-d35cf6d75469

Since it changes the parameters of the widget I thought "feat!" was correct ❓

### DIFF
--- a/lib/src/widgets/yaru_choice_chip_bar.dart
+++ b/lib/src/widgets/yaru_choice_chip_bar.dart
@@ -68,7 +68,6 @@ class _YaruChoiceChipBarState extends State<YaruChoiceChipBar> {
 
   @override
   Widget build(BuildContext context) {
-    final theme = Theme.of(context);
     final children = [
       for (int index = 0; index < widget.labels.length; index++)
         if (widget.isSelected[index])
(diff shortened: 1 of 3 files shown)


Jupi007:
Is it normal that the `listView` is visible below the `nextButton` but not below the `previousButton` in stacked mode?

jpnurmi:
I guess `feat:` is fine without an exclamation mark because the widget hasn't been released yet so the change is not breaking compared to a previous release

Feichtmeier:
ah! ok :)
will change this when I finalized the design. I have a better idea for the stack

Feichtmeier:
@Jupi007 @jpnurmi 

I thought, why show them at all if they are disable in the stack?

My questions would be just:
1) with or without a border
2) if they are round, make the chips here round as well (not everywhere)?


https://github.com/ubuntu/yaru_widgets.dart/assets/15329494/3fe43fd4-e162-4369-923b-634164a53120

Jupi007:
This look better :+1: 

I just have a few note:
- What about add a fade in/out on buttons?
- Maybe add a radius clip on the `listView` with the same radius as buttons, so chips border don't leak at corners.

Feichtmeier:
https://github.com/ubuntu/yaru_widgets.dart/assets/15329494/ae45feb5-fee6-46bc-bfa0-b347e02c7748

Clip was a good idea!
AnimatedContainer does not work, any other idea?

Jupi007:
> AnimatedContainer does not work, any other idea?

Why is it not working? 🤔

Feichtmeier:
No idea 🤷 

```dart
if (_enableGoPreviousButton)
              AnimatedContainer(
                duration: const Duration(seconds: 5),
                curve: Curves.bounceInOut,
                child: Positioned(
                  left: 0,
                  child: goPreviousButton,
                ),
              ),
              ```
              
              no animation 🤷 
is it the wrong place?

**Comment by Jupi007** [comment]
You need to use `AnimatedOpacity`, and always include the widget in the tree (drop `if (_enableGoPreviousButton)`).

**Comment by Feichtmeier** [comment]
the other way round but now it works thanks!

File: lib/src/widgets/yaru_choice_chip_bar.dart [hunk 1]
```diff
@@ -86,63 +86,140 @@ class _YaruChoiceChipBarState extends State<YaruChoiceChipBar> {
...
+                    _controller.position.pixels + widget.animationStep,
+                    duration: widget.duration,
+                    curve: Curves.bounceIn,
+                  )
+              : null,
+          child: const Icon(YaruIcons.go_next),
+        ),
+      ),
+    );
+
```
Jupi007, line 152: Those two buttons are basically the same except the icon and the animation position.
Maybe you can create a factory function at the same place in `build()` with the icon and the animation step as parameter?

Feichtmeier: I extracted the buttons but I think keeping the animated opacit in place of the build makes it more readble. Ok for you?

Jupi007: I agree :+1:

Jupi007 (approved):
LGTM :+1:

Feichtmeier:
Eventually I could add the option to parameterize the radius of the buttons and the chips because this may look better if everything was completely round

Feichtmeier:
@jpnurmi okay if I merge or anything you want to see changed in the design or code?

PR #727, opened 2023-06-30
chore(dependabot): configure semantic commits

jpnurmi:
I don't know how exactly dependabot decides between semantic vs. non-semantic commits but the first PR it opened didn't use semantic commits so try explicitly configuring it:

- #726


d-loose (approved):
LGTM :+1:

PR #730, opened 2023-07-04
feat: expose YaruNavigationPage.navigatorKey and onGenerateRoute

jpnurmi:
Allows pushing (named) routes onto YaruNavigationPage's internal Navigator. For example:

```dart
YaruNavigationPage(
  navigatorKey: _navigatorKey,
  ...
  onGenerateRoute: (settings) => switch (settings.name) {
    '/detail' => MaterialPageRoute(
        settings: settings,
        builder: (_) => DetailPage(settings.arguments as Detail),
      ),
    '/search' => MaterialPageRoute(
        settings: settings,
        builder: (_) => SearchPage(query: settings.arguments as String),
      ),
    _ => null,
  },
)
```

From within the navigation page:
```dart
Navigator.pushNamed(context, '/detail', arguments: detail);
```

From anywhere else such as a global search bar:
```dart
_navigatorKey.currentState!.pushNamed('/search', arguments: query);
```

Issue #512, opened 2023-01-13
YaruMasterDetailPage: Navigator state is lost on layout switch


d-loose:
https://user-images.githubusercontent.com/113362648/212308916-c3b4adf0-e13d-4f5a-8a3c-d41ff0e734a7.mp4

When using the `Navigator`s provided by `YaruLandscapeLayout` and `YaruPortraitLayout` their state is lost when the layout switches between them.

Feichtmeier:
I think the problem is the layout builder. Maybe we could provide the landscape layout without the portrait layout if we don't fix the layout builder problem

jpnurmi:
I don't know what I'm doing wrong here but in theory, something like this should work. This retains the state but somehow breaks Navigator after relayouting/reparenting it back and forth.

[Screencast from 2023-01-15 21-16-22.webm](https://user-images.githubusercontent.com/140617/212564773-c4a1fb6a-8b18-4689-99c2-b0548201e0f6.webm)

```dart
import 'package:flutter/foundation.dart';
import 'package:flutter/material.dart';

const kColors = [Colors.red, Colors.green, Colors.blue, Colors.yellow];

void main() => runApp(MaterialApp(home: MasterDetailPage()));

class MasterDetailPage extends StatefulWidget {
  MasterDetailPage({super.key});

  @override
  _MasterDetailPageState createState() => _MasterDetailPageState();
}

class _MasterDetailPageState extends State<MasterDetailPage> {
  final _navigatorKey = GlobalKey<NavigatorState>();
  var index = 0;

  @override
  Widget build(BuildContext context) {
    return OrientationBuilder(builder: (context, orientation) {
      final master = MasterListView(
        length: kColors.length,
        selected: index,
        onTap: (index) => setState(() => this.index = index),
      );

      final navigator = Navigator(
        key: _navigatorKey,
        pages: [
          if (orientation == Orientation.portrait)
            MaterialPage(
              key: ValueKey(-1),
              child: master,
            ),
          MaterialPage(
            key: ValueKey(index),
            child: DetailPage(color: kColors[index % kColors.length]),
          ),
        ],
        onPopPage: (route, result) => route.didPop(result),
      );

      return Scaffold(
        body: orientation == Orientation.landscape
            ? Row(
                children: [
                  SizedBox(width: 200, child: master),
                  Expanded(child: navigator),
                ],
              )
            : navigator,
      );
    });
  }
}

class MasterListView extends StatelessWidget {
[...]

PR #732, opened 2023-07-12
feat: expose NavigationPage.onPopPage

jpnurmi:
Continues where #730 and #731 left off. Exposing `Navigator.onPopPage` allows updating the GUI state when a dynamically pushed (e.g. search or detail) page is popped off.


Feichtmeier:
Uhhhh this sounds nice. Does this also help with preservation of the state of a page if the app is rebuilt via layout/orientation builders?

jpnurmi:
Do you mean this?
- #512

That should be fixable now that the [attached example](https://github.com/ubuntu/yaru_widgets.dart/issues/512#issuecomment-1383241996) works fine with the latest stable Flutter but this PR is about something else i.e. being able to observe from the outside of NavigationPage what happens in its internal Navigator.

Feichtmeier:
Okay that's also nice ok 👍

d-loose (approved):
LGTM :+1:

PR #734, opened 2023-07-20
feat: add YaruSearchField and YaruSearchTitleField

Feichtmeier:
I thought it is time to find a sane default for searchfields and searchfields inside titlebars since this seems to become a common pattern in our apps.

This does not need to be the final result and should eventually be parameterized more but this could be a good start.

And in general with flutter, aynthing is changeable.

Opinions @madsrh @Jupi007 @jpnurmi @elioqoshi ?

| SearchActive | SearchNotActive |
|--------|-------|
|Bildschirmfoto 2023-07-20 um 14 14 11|Bildschirmfoto 2023-07-20 um 14 14 23|
|Bildschirmfoto 2023-07-20 um 14 14 45|Bildschirmfoto 2023-07-20 um 14 15 01 |

### DIFF
--- a/lib/constants.dart
+++ b/lib/constants.dart
@@ -16,3 +16,8 @@ const kYaruButtonRadius = 6.0;
 /// The default breakpoint width [YaruMasterDetailPage] uses for switching
 /// between portrait and landscape modes.
 const kYaruMasterDetailBreakpoint = 620.0;
+
+/// The best height for any item inside a [YaruWindowTitleBar]
+const kYaruWindowTitleBarItemHeight = 35.0;
+
+const kYaruIconSize = 16.0;
--- a/lib/constants.dart
+++ b/lib/constants.dart
@@ -20,4 +20,5 @@ const kYaruMasterDetailBreakpoint = 620.0;
 /// The best height for any item inside a [YaruWindowTitleBar]
 const kYaruWindowTitleBarItemHeight = 35.0;
 
+/// The default icon size
 const kYaruIconSize = 16.0;
--- a/lib/constants.dart
+++ b/lib/constants.dart
@@ -17,8 +17,8 @@ const kYaruButtonRadius = 6.0;
 /// between portrait and landscape modes.
 const kYaruMasterDetailBreakpoint = 620.0;
 
-/// The best height for any item inside a [YaruWindowTitleBar]
-const kYaruWindowTitleBarItemHeight = 35.0;
+/// The best height for any item inside a [YaruTitleBar]
+const kYaruTitleBarItemHeight = 35.0;
 
 /// The default icon size
 const kYaruIconSize = 16.0;
--- a/lib/src/widgets/yaru_search_field.dart
+++ b/lib/src/widgets/yaru_search_field.dart
@@ -143,7 +143,7 @@ class _YaruSearchFieldState extends State<YaruSearchField> {
                 (widget.fillColor ?? theme.dividerColor).scale(lightness: 0.1),
             suffixIconConstraints:
                 const BoxConstraints(maxWidth: kYaruTitleBarItemHeight),
-            suffixIcon: widget.onClear == null
+            suffixIcon: widget.onClear == null || _controller.text.isEmpty
                 ? null
                 : IconButton(
                     style: IconButton.styleFrom(
@@ -231,7 +231,7 @@ class _YaruSearchTitleFieldState extends State<YaruSearchTitleField> {
       width: widget.width,
       child: ClipRRect(
         borderRadius: BorderRadius.all(widget.radius),
-        clipBehavior: Clip.antiAliasWithSaveLayer,
+        clipBehavior: Clip.antiAlias,
         child: Stack(
           alignment: Alignment.centerLeft,
           children: [
@@ -240,6 +240,7 @@ class _YaruSearchTitleFieldState extends State<YaruSearchTitleField> {
                 child: SizedBox(
                   height: kYaruTitleBarItemHeight,
                   child: YaruSearchField(
+                    text: widget.text,
                     style: widget.style,
                     radius: widget.radius,
                     height: widget.width,
--- a/lib/src/widgets/yaru_search_field.dart
+++ b/lib/src/widgets/yaru_search_field.dart
@@ -13,16 +13,24 @@ class YaruSearchField extends StatefulWidget {
     this.hintText,
     this.onSearchActive,
     this.height = kYaruWindowTitleBarItemHeight,
-    this.contentPadding =
-        const EdgeInsets.only(bottom: 10, top: 10, right: 15, left: 45),
+    this.contentPadding = const EdgeInsets.only(
+      bottom: 10,
+      top: 10,
+      right: 15,
+      left: 15,
+    ),
+    this.autofocus = true,
+    this.onClear,
   });
 
   final String? text;
   final String? hintText;
   final void Function(String? value)? onSubmitted;
+  final void Function()? onClear;
   final void Function()? onSearchActive;
   final double height;
   final EdgeInsets contentPadding;
+  final bool autofocus;
 
   @override
   State<YaruSearchField> createState() => _YaruSearchFieldState();
@@ -53,17 +61,23 @@ class _YaruSearchFieldState extends State<YaruSearchField> {
       borderRadius: BorderRadius.circular(100),
     );
 
+    const suffixRadius = BorderRadius.only(
+      topRight: Radius.circular(kYaruWindowTitleBarItemHeight),
+      bottomRight: Radius.circular(
+        kYaruWindowTitleBarItemHeight,
+      ),
+    );
     return KeyboardListener(
       focusNode: focusNode,
       onKeyEvent: (value) {
         if (value.logicalKey == LogicalKeyboardKey.escape) {
-          widget.onSearchActive?.call();
+          _clear();
         }
       },
       child: SizedBox(
         height: widget.height,
         child: TextField(
-          autofocus: true,
+          autofocus: widget.autofocus,
           style: theme.textTheme.bodyMedium,
           strutStyle: const StrutStyle(
             leading: 0.2,
@@ -81,10 +95,31 @@ class _YaruSearchFieldState extends State<YaruSearchField> {
             hintText: widget.hintText,
             fillColor: theme.dividerColor,
             hoverColor: theme.dividerColor.scale(lightness: 0.1),
+            suffixIcon: widget.onClear == null
+                ? null
+                : IconButton(
+                    style: IconButton.styleFrom(
+                      shape: const RoundedRectangleBorder(
+                        borderRadius: suffixRadius,
+                      ),
+                    ),
+                    onPressed: _clear,
+                    icon: const ClipRRect(
+                      borderRadius: suffixRadius,
+                      child: Icon(
+                        YaruIcons.edit_clear,
+                      ),
+                    ),
+                  ),
           ),
         ),
       ),
     );
+  }
+
+  void _clear() {
+    widget.onClear?.call();
+    _controller.clear();
   }
 }
 
@@ -95,12 +130,24 @@ class YaruSearchFieldTitle extends StatefulWidget {
     required this.title,
     this.width = 190,
     this.titlePadding = const EdgeInsets.only(left: 45.0),
+    this.autofocus = true,
+    this.text,
+    this.hintText,
+    this.onSubmitted,
+    this.onClear,
+    this.onSearchActive,
   });
 
   final bool searchActive;
   final Widget title;
   final double width;
   final EdgeInsets titlePadding;
+  final bool autofocus;
+  final String? text;
+  final String? hintText;
+  final void Function(String? value)? onSubmitted;
+  final void Function()? onClear;
+  final void Function()? onSearchActive;
 
   @override
   State<YaruSearchFieldTitle> createState() => _YaruSearchFieldTitleState();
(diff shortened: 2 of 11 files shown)


jpnurmi:
Did it somehow stretch the title bar or does it just look extra tall because of the macOS window controls?

Feichtmeier:
I think I just hit CMD+ + and then in vscode even the flutter window increases its titlebar :)

this is how it looks on windows
grafik

the way the example is built I can't just put a scaffold as a page. its possible but I would need to edit aaaaaall page item pages which is kind of overkill for this PR

Jupi007:
I really like it!

madsrh:
> I thought it is time to find a sane default for searchfields and searchfields inside titlebars since this seems to become a common pattern in our apps.

💪  Sounds like a good idea

Looks good but TBH I'm not a fan of the two background colors. Could you explain again why you want the icon to have a different background color? Is it clickable? I'm sorry this was mentioned elsewhere

Feichtmeier:
yes its an icon button. when you click it the searchbar is layed under it (stack)

madsrh:
> yes its an icon button. when you click it the searchbar is layed under it (stack)

Makes sense. I couldn't tell from the screenshot 😅

Feichtmeier:
[Bildschirmaufzeichnung vom 2023-07-24, 17-38-14.webm](https://github.com/ubuntu/yaru_widgets.dart/assets/15329494/545f7ebd-64a7-4c87-8aed-45ebb657c0dd)

ok I think this is pretty good
yet modular enough I hope

if there is an onClear callback provided the suffix icon shows up with a clear button

the search field and button can be used either combiend in the title or separated

jpnurmi:
I understand well why, but this looks mis-aligned when the search is not active. :)

image

lib/src/widgets/yaru_search_field.dart
```diff
@@ -0,0 +1,267 @@
...
+      width: widget.width,
+      child: Stack(
+        alignment: Alignment.centerLeft,
+        children: [
+          if (_searchActive)
+            Center(
+              child: SizedBox(
+                height: kYaruWindowTitleBarItemHeight,
+                child: YaruSearchField(
+                  height: widget.width,
```
jpnurmi, line 199: height: width? is this correct?

Feichtmeier:
yeah the example is stupid here because you would prbly not change the title in the text field 😸 
I'll see if I can improve or parameterize the space between the title and the search icon

lib/src/widgets/yaru_search_field.dart
```diff
@@ -0,0 +1,267 @@
...
+  State<YaruSearchFieldTitle> createState() => _YaruSearchFieldTitleState();
+}
+
+class _YaruSearchFieldTitleState extends State<YaruSearchFieldTitle> {
+  late bool _searchActive;
+
+  @override
+  void initState() {
+    super.initState();
+    _searchActive = widget.searchActive;
```
jpnurmi, line 184: same in `didUpdateWidget()`?

lib/src/widgets/yaru_search_field.dart
```diff
@@ -0,0 +1,267 @@
...
+      ),
+    );
+  }
+}
+
+/// A pre-styled [YaruIconButton], ideally used in combination with [YaruSearchField]
+class YaruSearchButton extends StatelessWidget {
+  const YaruSearchButton({
+    super.key,
+    required this.searchActive,
```
jpnurmi, line 233: required makes sense for the search field but what if the button is used standalone?

lib/src/widgets/yaru_search_field.dart
```diff
@@ -0,0 +1,267 @@
...
+      focusNode: focusNode,
+      onKeyEvent: (value) {
+        if (value.logicalKey == LogicalKeyboardKey.escape) {
+          _clear();
+        }
+      },
+      child: SizedBox(
+        height: widget.height,
+        child: TextField(
+          autofocus: widget.autoFocus,
```
jpnurmi, line 98: nit-pick: Flutter doesn't capitalize Focus in autofocus

lib/src/widgets/yaru_search_field.dart
```diff
@@ -0,0 +1,267 @@
...
+  @override
+  void dispose() {
+    _controller.dispose();
+    super.dispose();
+  }
+
+  @override
+  Widget build(BuildContext context) {
+    final theme = Theme.of(context);
+    final focusNode = FocusNode();
```
jpnurmi, line 75: create as member of _YaruSearchFieldState and dispose in dispose()

lib/src/widgets/yaru_search_field.dart
```diff
@@ -0,0 +1,267 @@
...
+  final Widget title;
+  final double width;
+  final EdgeInsets titlePadding;
+  final bool autoFocus;
+  final String? text;
+  final String? hintText;
+  final void Function(String? value)? onSubmitted;
+  final void Function(String)? onChanged;
+  final void Function()? onClear;
+  final void Function()? onSearchActive;
```
jpnurmi, line 172: missing implementation?

lib/constants.dart
```diff
@@ -16,3 +16,9 @@ const kYaruButtonRadius = 6.0;
 /// The default breakpoint width [YaruMasterDetailPage] uses for switching
 /// between portrait and landscape modes.
 const kYaruMasterDetailBreakpoint = 620.0;
+
+/// The best height for any item inside a [YaruWindowTitleBar]
+const kYaruWindowTitleBarItemHeight = 35.0;
```
jpnurmi, line 21: possibly applies to any YaruTitleBar, not just YaruWindowTitleBar?

Feichtmeier:
Thanks @jpnurmi there were indeed some heavy issues :see_no_evil:

jpnurmi:
A couple of remarks, feel free to ignore them if you think they are not relevant.

- There's something that makes it feel somehow... unlike Yaru. Perhaps the fully rounded corners? It's not necessarily bad, but no other search field looks like this on my desktop. :)
- The name "search field title" feels somehow backward. Would "search title field" sound more natural?

Feichtmeier:
Well, I could just add a border radius parameter and you guys choose what you like. 
Personally I feel that those widgets need an iconic look because they fulfill a special purpose. They're not regular Text fields like in forms. (Not saying forms are not important)

I'll change the name! 😎👌

Feichtmeier:
@jpnurmi renamed as requested plus added a radius parameter
you might ask :thinking: "why not BorderRadius?"
because the suffix icon has only the right edges rounded :)

image
example with kYaruButtonRadius from yaru.dart

jpnurmi:
The search icon button is 35x35px whereas the clear icon button is 40x35px:

image

The clear button can be clicked even if the search field is empty. It's not immediately obvious that you need to click the search button again to close the search field. Perhaps the clear button could close it if the field is already empty?

[Screencast from 2023-07-26 08-18-49.webm](https://github.com/ubuntu/yaru_widgets.dart/assets/140617/2de96ff6-5485-4fff-87e4-3943aca8f7e4)

Feichtmeier:
> The search icon button is 35x35px whereas the clear icon button is 40x35px:
> 
> image
> 
> The clear button can be clicked even if the search field is empty. It's not immediately obvious that you need to click the search button again to close the search field. Perhaps the clear button could close the field if it's already empty?
> Screencast.from.2023-07-26.08-18-49.webm

I would leave the logic when to enable the clear button via the callback to the user of the widget 

I'll fix the clear button width

Feichtmeier:
@jpnurmi nurmi

fixed width of the clear button

plus added a YaruSearchFieldStyle
filled, outlined, filledOutlined parameters
plus filledcolor and bordercolor (optional)

filled
grafik

outlined
grafik

filledoutlined
grafik

anything else?

lib/src/widgets/yaru_search_field.dart
```diff
@@ -0,0 +1,351 @@
...
+        borderRadius: BorderRadius.all(widget.radius),
+        clipBehavior: Clip.antiAliasWithSaveLayer,
+        child: Stack(
+          alignment: Alignment.centerLeft,
+          children: [
+            if (_searchActive)
+              Center(
+                child: SizedBox(
+                  height: kYaruTitleBarItemHeight,
+                  child: YaruSearchField(
```
jpnurmi, line 242: missing `text: widget.text`

lib/src/widgets/yaru_search_field.dart
```diff
@@ -0,0 +1,351 @@
...
+    _searchActive = widget.searchActive;
+  }
+
+  @override
+  Widget build(BuildContext context) {
+    return SizedBox(
+      width: widget.width,
+      child: ClipRRect(
+        borderRadius: BorderRadius.all(widget.radius),
+        clipBehavior: Clip.antiAliasWithSaveLayer,
```
jpnurmi, line 234: why `Clip.antiAliasWithSaveLayer`? as the docs say, it's much slower than `Clip.antiAlias` and should rarely be used.

lib/src/widgets/yaru_search_field.dart
```diff
@@ -0,0 +1,351 @@
...
+  /// it will fall back to `ColorScheme.dividerColor`.
+  final Color? fillColor;
+
+  @override
+  State<YaruSearchField> createState() => _YaruSearchFieldState();
+}
+
+class _YaruSearchFieldState extends State<YaruSearchField> {
+  late TextEditingController _controller;
+  final _focusNode = FocusNode();
```
jpnurmi, line 76: i have a strong feeling we should allow users to pass both or else they won't be able to control selection or focus

jpnurmi (commented):
I still think the clear button should close an empty search field or at least become disabled if it doesn't have any effect. Also, the search field is left open forever even when it's empty and loses focus. It's not immediately obvious how to get rid of the field and return back to the normal title. Especially since the clear button is confusingly enabled yet does nothing... :) Should either YaruSearch(Title)Field or the example handle these cases?

lib/src/widgets/yaru_search_field.dart
Feichtmeier: yes you are right, there is something f*cked up with the focus on rebuild

Feichtmeier: you think that those fields should be completely out here and make the field totally stateless?

Feichtmeier: hiding the clear button now if _controller.text is empty

jpnurmi: they are usually only needed for more advanced use cases. it might be inconvenient to always require them. i'd follow the common practice of creating and disposing them here if the user didn't supply them.

Feichtmeier: ok!

Feichtmeier: like this?

lib/src/widgets/yaru_search_field.dart
```diff
@@ -0,0 +1,372 @@
...
+  /// Optional [FocusNode] for the internal [KeyboardListener]
+  final FocusNode? focusNode;
+
+  @override
+  State<YaruSearchField> createState() => _YaruSearchFieldState();
+}
+
+class _YaruSearchFieldState extends State<YaruSearchField> {
+  late TextEditingController? _controller;
+  late FocusNode? _focusNode;
```
jpnurmi, line 84:
```suggestion
  late TextEditingController _controller;
  late FocusNode _focusNode;
```

lib/src/widgets/yaru_search_field.dart
```diff
@@ -0,0 +1,372 @@
...
+
+  @override
+  void initState() {
+    super.initState();
+    _controller = widget.controller ?? TextEditingController(text: widget.text);
+    _focusNode = widget.focusNode;
```
jpnurmi, line 90:
```suggestion
    _focusNode = widget.focusNode ?? FocusNode();
```

lib/src/widgets/yaru_search_field.dart
```diff
@@ -0,0 +1,372 @@
...
+  }
+
+  @override
+  void dispose() {
+    _controller?.dispose();
+    _focusNode?.dispose();
```
jpnurmi, line 96:
```suggestion
    if (widget.controller == null) _controller.dispose();
    if (widget.focusNode == null) _focusNode.dispose();
```

lib/src/widgets/yaru_search_field.dart
```diff
@@ -0,0 +1,372 @@
...
+      borderRadius: BorderRadius.all(widget.radius),
+    );
+
+    final suffixRadius = BorderRadius.only(
+      topRight: widget.radius,
+      bottomRight: widget.radius,
+    );
+
+    return KeyboardListener(
+      focusNode: _focusNode ?? FocusNode(),
```
jpnurmi, line 123:
```suggestion
      focusNode: _focusNode,
```

lib/src/widgets/yaru_search_field.dart
```diff
@@ -0,0 +1,372 @@
...
+  late TextEditingController? _controller;
+  late FocusNode? _focusNode;
+
+  @override
+  void initState() {
+    super.initState();
+    _controller = widget.controller ?? TextEditingController(text: widget.text);
+    _focusNode = widget.focusNode;
+  }
+
```
jpnurmi, line 92: the usual rebuild checks, something like
```dart
  @override
  void didUpdateWidget(YaruSearchField oldWidget) {
    super.didUpdateWidget(oldWidget);

    if (widget.controller != oldWidget.controller) {
      if (oldWidget.controller == null) _controller.dispose();
      _controller = widget.controller ?? TextEditingController(text: widget.text);
    }
  }
```

lib/src/widgets/yaru_search_field.dart
jpnurmi: what could possibly go wrong when suggesting changes written on the fly via github's web GUI :)

lib/src/widgets/yaru_search_field.dart
Feichtmeier: thanks, also for focusnode?

lib/src/widgets/yaru_search_field.dart
Feichtmeier: haha :laughing: well fixed it locally now :smile:

lib/src/widgets/yaru_search_field.dart
jpnurmi: oh yes that too :+1:

Feichtmeier: okay did this

Feichtmeier:
@jpnurmi can I merge?

jpnurmi (approved):
LGTM :+1:

PR #737, opened 2023-07-25
feat(master-detail): expose breakpoint as widget property

jpnurmi:
And document that `0` and `double.infinity` can be used as breakpoints to force landscape and portrait layouts, respectively.

### DIFF
--- a/lib/src/widgets/master_detail/yaru_master_detail_page.dart
+++ b/lib/src/widgets/master_detail/yaru_master_detail_page.dart
@@ -58,6 +58,7 @@ class YaruMasterDetailPage extends StatefulWidget {
     this.emptyBuilder,
     this.layoutDelegate =
         const YaruMasterFixedPaneDelegate(paneWidth: _kDefaultPaneWidth),
+    this.breakpoint,
     this.appBar,
     this.appBarBuilder,
     this.bottomBar,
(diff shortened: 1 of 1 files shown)


d-loose (approved):
LGTM :+1:

Feichtmeier (approved):
Awesome! Can you read my mind? XD 😂 that's really useful

Feichtmeier:
@jpnurmi one of the past YMD PRs breaks musicpod with this error
Calculating the intrinsic dimensions would require running the layout callback speculatively, which might mutate the live render object tree.
The relevant error-causing widget was
YaruMasterDetailPage
app.dart:326
When the exception was thrown, this was the stack

any idea what I need to adapt now?

jpnurmi:
@Feichtmeier I doubt it's specific to this PR because it only exposed an optional property that defaults to the theme value as before. Anyway, I'd try removing the [kung fu](https://github.com/ubuntu/yaru_widgets.dart/pull/736#issuecomment-1648206055) and if the problem persists, please minimize it to a test case.

Feichtmeier:
I found the "issue"

paul added a responsive master tile
https://github.com/ubuntu-flutter-community/musicpod/blob/main/lib/app/responsive_master_tile.dart
which uses a layout builder, replacing them with mastertiles makes the error go away, however not the mastertiles are not responsive anymore 😆 
in musicpod the sidebar shrinks to only show the icons (kinda like a navi rail) when the window is too narrow
thats why paul was so nice and added those responsive tiles
do you think you could implement such behavor in YMD?

jpnurmi:
Was the `availableWidth` argument added for this use case?

- https://github.com/ubuntu/yaru_widgets.dart/pull/708

Feichtmeier:
maybe, CC @Jupi007

Feichtmeier:
Okay guys I will just switch to the indented design for musicpod for the moment
This switching of delegates dep on window width is too fragile and also rebuilds the page

This looks also fine :+1: have a nice eve :night_with_stars: 
[Bildschirmaufzeichnung vom 2023-07-26, 21-15-34.webm](https://github.com/ubuntu/yaru_widgets.dart/assets/15329494/cf6ce09f-bb6c-4269-96c3-a6c480e8c408)

PR #738, opened 2023-07-27
fix(YaruMasterTile): let onTap override scope ontap

### LINKED ISSUE #718: Add a navigation focused alternative to YaruMasterDetailPage
Currently you select a tile in the master view and the detail view is completely rebuild.
This is fine for apps like settings where each page is somehow detached from the rest.

But when you want to navigate more freely it is hard to code around YaruMasterDetailPage

Proposal:

instead of giving the master tiles so much "power" I would suggest to create an alternative widget that keeps the master view but does not change the detail view depending on the index.
The index is still needed to tell a tile if it is selected or not but that's it.
The details view should be just the navigator. And we can set an ontap callback for the master tiles which push to a new material page.
This would also allow an easy forward and back navigation as for example in spotify.

This should be done inside another Widget and should not replace YaruMasterDetailPage.

Too late for mock-ups. I hope my description was clear if not I can add mock-ups in the next week.

Feichtmeier:
~~This adds a parameter to disable looking up the YaruMasterTileScope for tiles added to the list inside YMD that should not open a page.~~

~~Is this the correct approach @jpnurmi ?~~

Use case: 1 tile inside the list should open a dialog and not select/load a page

Closes #718

### DIFF
--- a/lib/src/widgets/master_detail/yaru_master_tile.dart
+++ b/lib/src/widgets/master_detail/yaru_master_tile.dart
@@ -35,13 +35,15 @@ class YaruMasterTile extends StatelessWidget {
   /// See [ListTile.trailing].
   final Widget? trailing;
 
-  /// See [ListTile.onTap].
+  /// An optional [VoidCallback] forwarded to the internal [ListTile]
+  /// If not provided [YaruMasterTileScope] `onTap` will be called.
   final VoidCallback? onTap;
 
   @override
   Widget build(BuildContext context) {
     final theme = Theme.of(context);
     final scope = YaruMasterTileScope.maybeOf(context);
+
     final isSelected = selected ?? scope?.selected ?? false;
     final scrollbarThicknessWithTrack =
         _calcScrollbarThicknessWithTrack(context);
@@ -76,8 +78,11 @@ class YaruMasterTile extends StatelessWidget {
           trailing: trailing,
           selected: isSelected,
           onTap: () {
-            scope?.onTap();
-            onTap?.call();
+            if (onTap != null) {
+              onTap!.call();
+            } else {
+              scope?.onTap();
+            }
           },
         ),
       ),


jpnurmi:
What's the problem you're trying to solve? `YaruMasterTileScope.maybeOf()` returns null if it's not wrapped into a scope.

Feichtmeier:
I think you commented 2 secs before I updated the PR desciption :P 

This is the use case, X tiles inside the list should not open a page but just onTap:()=> doSomething()

[Bildschirmaufzeichnung vom 2023-07-27, 13-41-28.webm](https://github.com/ubuntu/yaru_widgets.dart/assets/15329494/a87990ef-5682-4a74-9ac2-69559d6beae7)

jpnurmi:
What if we'd allow `YaruMasterTile.onTap()` "override" the scope i.e. if you specify your own `onTap` handler then the scope handler won't be called?

```diff
diff --git a/lib/src/widgets/master_detail/yaru_master_tile.dart b/lib/src/widgets/master_detail/yaru_master_tile.dart
index f655985..5ba8a32 100644
--- a/lib/src/widgets/master_detail/yaru_master_tile.dart
+++ b/lib/src/widgets/master_detail/yaru_master_tile.dart
@@ -76,8 +76,11 @@ class YaruMasterTile extends StatelessWidget {
           trailing: trailing,
           selected: isSelected,
           onTap: () {
-            scope?.onTap();
-            onTap?.call();
+            if (onTap != null) {
+              onTap!.call();
+            } else {
+              scope?.onTap();
+            }
           },
         ),
       ),
```

Feichtmeier:
> What if we'd allow `YaruMasterTile.onTap()` "override" the scope i.e. if you specify your own `onTap` handler then the scope handler won't be called?
> 
> ```diff
> diff --git a/lib/src/widgets/master_detail/yaru_master_tile.dart b/lib/src/widgets/master_detail/yaru_master_tile.dart
> index f655985..5ba8a32 100644
> --- a/lib/src/widgets/master_detail/yaru_master_tile.dart
> +++ b/lib/src/widgets/master_detail/yaru_master_tile.dart
> @@ -76,8 +76,11 @@ class YaruMasterTile extends StatelessWidget {
>            trailing: trailing,
>            selected: isSelected,
>            onTap: () {
> -            scope?.onTap();
> -            onTap?.call();
> +            if (onTap != null) {
> +              onTap!.call();
> +            } else {
> +              scope?.onTap();
> +            }
>            },
>          ),
>        ),
> ```

yeah that would be even better, that was funnily what I assumed would happen when you override ontap :smile: 

this would also fix https://github.com/ubuntu/yaru_widgets.dart/issues/718 btw

jpnurmi (approved):
LGTM :+1:

PR #740, opened 2023-07-29
fix: add the missing optional icon parameters

### LINKED ISSUE #739: add the missing optional icon parameters
@Jupi007 @jpnurmi 

IMO we should remove the dependency to yaru icons
The amount of places where we use them is so low that I don't see why we just shouldn't add a required IconData or Icon parameter to these widgets:

grafik

Apart from making the dependency management easier across our apps and libs this would also make yaru widgets to be more opened to be used for example with adwaita icons or material icons

Feichtmeier:
Fixes #739

### DIFF
--- a/lib/src/widgets/yaru_back_button.dart
+++ b/lib/src/widgets/yaru_back_button.dart
@@ -11,6 +11,7 @@ class YaruBackButton extends StatelessWidget {
     super.key,
     this.onPressed,
     this.style,
+    this.icon,
   });
 
   /// An optional callback that is called when the button is pressed.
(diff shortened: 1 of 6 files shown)


Feichtmeier:
@jpnurmi any veto for setting the branch protection rule "require 1 approval before merging"?

Jupi007 (approved):
LGTM 👍

Jupi007:
Shouldn't we also add the icon properties to the theme classes ?

Feichtmeier:
> Shouldn't we also add the icon properties to the theme classes ?

if you have material themes as blueprint no
there is no material theme that presets a prop to a certain Widget
only things like themes, sizes and colors

Jupi007:
> if you have material themes as blueprint no
> there is no material theme that presets a prop to a certain Widget
> only things like themes, sizes and colors

Indeed, you're right.

PR #743, opened 2023-07-29
chore: relax yaru_icons.dart version constraint

### LINKED ISSUE #742: relax yaru_icons.dart version constraint
What about relaxing the version constraint?

Before:
```yaml
yaru_icons: ^1.0.0 # same as '>=1.0.0 <2.0.0'
```
After:
```yaml
yaru_icons: '>=1.0.0'
```

_Originally posted by @jpnurmi in https://github.com/ubuntu/yaru_widgets.dart/issues/739#issuecomment-1656714020_

Feichtmeier:
Fixes #742

### DIFF
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -12,7 +12,7 @@ dependencies:
   flutter:
     sdk: flutter
   yaru: ^0.9.0
-  yaru_icons: ^1.0.0
+  yaru_icons: ">=1.0.0"
   yaru_window: ^0.1.1
 
 dev_dependencies:


jpnurmi:
Hmm, pub.dev seems to throw a warning :(

> Validating package...
> Package validation found the following potential issue:
> * Your dependency on "yaru_icons" should have an upper bound. For example:
>   
>   dependencies:
>     yaru_icons: ^1.0.0
>   
>   Without an upper bound, you're promising to support all future versions of yaru_icons.
> 
> Package has 1 warning.

pubspec.yaml
```diff
@@ -12,7 +12,7 @@ dependencies:
   flutter:
     sdk: flutter
   yaru: ^0.9.0
-  yaru_icons: ^1.0.0
+  yaru_icons: ">=1.0.0"
```
jpnurmi, line 15: my bad, i didn't realize pub.dev being so cranky about this. what about ">=1.0.0 <3.0.0" to make pub.dev happy?

PR #746, opened 2023-08-05
chore!: update yaru icons

Feichtmeier:

### DIFF
--- a/example/lib/code_snippet_button.dart
+++ b/example/lib/code_snippet_button.dart
@@ -78,7 +78,8 @@ class _CodeDialog extends StatelessWidget {
               child: Column(
                 mainAxisSize: MainAxisSize.min,
                 children: [
-                  YaruAnimatedNoNetworkIcon(
+                  YaruAnimatedIcon(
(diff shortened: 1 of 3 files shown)


example/lib/code_snippet_button.dart
```diff
@@ -78,7 +78,8 @@ class _CodeDialog extends StatelessWidget {
               child: Column(
                 mainAxisSize: MainAxisSize.min,
                 children: [
-                  YaruAnimatedNoNetworkIcon(
+                  YaruAnimatedIcon(
+                    const YaruAnimatedOkIcon(),
```
Jupi007, line 82: I think you should use                   `YaruAnimatedNoNetworkIcon` here 🙂

Jupi007 (changes requested):

example/lib/code_snippet_button.dart
Feichtmeier: woops

Jupi007 (approved):
LGTM :rocket:

PR #748, opened 2023-08-07
fix: relax yaru version constraint; update example

Feichtmeier:
This changes the look of yaru widgets because of the yaru color changes
grafik

I think we can now have a brighter sidebar like in musicpod, maybe with an option to hide the divider
grafik


jpnurmi (approved):
LGTM :+1:

PR #750, opened 2023-08-07
feat: change sidebar color, improve YMD theme

Feichtmeier:
This brightens up the sidebar for the dark theme a bit and darkens up the sidebar for the light theme a bit. This improves the visual focus on the content on the right (or left)
Plus this also adapts our looks to libadwaita 

Possible variants:

grafik

grafik

**Default:**

grafik

light:

grafik

**default:**
grafik


Feichtmeier:
those variants are not inherited by some enum or so because they are too complex for 
appbar | appbar
list | detailpage
combos

but they are possible by setting the background colors of YaruWindowTitleBars and by wrapping the YMD in a YMD with copy with and changing the defaults

Feichtmeier:
@jpnurmi any veto? :smile:

jpnurmi (approved):
LGTM :+1: 

p.s. navi-rail too?

Feichtmeier:
> LGTM 👍
> 
> p.s. navi-rail too?

yeah I should, right?
in this PR or another?
Navirail in material defaults to colorscheme.surface
I wonder if we could rather let yaru navi rail use the material theme? because it is so close to it

Jupi007:
>  I wonder if we could rather let yaru navi rail use the material theme? because it is so close to it

IMO it is not a good idea.

jpnurmi:
A separate PR is fine, of course, if you decide to apply similar changes. :)

Feichtmeier:
> > I wonder if we could rather let yaru navi rail use the material theme? because it is so close to it
> 
> IMO it is not a good idea.

okay then I apply those changes to YaruNavigationPageThemeData in a second PR

PR #383, opened 2022-11-14
Fix goldens images size (perfect fit)

Jupi007:
## Pull request checklist

- [x] Either this PR does not introduce visual changes

Fix goldens tests images size to perfectly fit the related widget.

PR #763, opened 2023-08-15
feat!: adapt to yaru compact sizes

Feichtmeier:
WIP until we have yaru 1.0.0 for pubspec

- 34 pixel height icons
- 20 is the default icon size now which affects also listtiles and buttons ofc

See https://github.com/ubuntu/yaru.dart/pull/383

@Jupi007 should we export yaru.dart constants instead and use those? or do we want to play safe and just make sure we sync them?


Jupi007:
> @Jupi007 should we export yaru.dart constants instead and use those? or do we want to play safe and just make sure we sync them?

We can export a restricted part of constants in a `public_constants.dart` file.

Feichtmeier:
I don't mind :smile: what would you prefer

Jupi007:
I would prefer to use constants from yaru.dart :)
But let's not export everything.

Feichtmeier:
snapshot, this already looks like a very big improvement to me
grafik

Jupi007 (approved):
LGTM :+1:

PR #768, opened 2023-08-18
fix: update to flutter 3.13

Feichtmeier:

### DIFF
--- a/lib/src/widgets/yaru_toggle_button_layout.dart
+++ b/lib/src/widgets/yaru_toggle_button_layout.dart
@@ -31,7 +31,10 @@ part of 'yaru_toggle_button.dart';
 enum _YaruToggleButtonSlot { leading, title, subtitle }
 
 class _YaruToggleButtonLayout extends RenderObjectWidget
-    with SlottedMultiChildRenderObjectWidgetMixin<_YaruToggleButtonSlot> {
+    with
+        // ignore: deprecated_member_use
+        SlottedMultiChildRenderObjectWidgetMixin<_YaruToggleButtonSlot,
+            RenderBox> {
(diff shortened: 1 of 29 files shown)


Feichtmeier:
please dont delete branch when you approved :P (using it in musicpod for the moment)

lib/src/widgets/yaru_toggle_button_layout.dart
```diff
@@ -31,7 +31,10 @@ part of 'yaru_toggle_button.dart';
 enum _YaruToggleButtonSlot { leading, title, subtitle }
 
 class _YaruToggleButtonLayout extends RenderObjectWidget
-    with SlottedMultiChildRenderObjectWidgetMixin<_YaruToggleButtonSlot> {
+    with
+        // ignore: deprecated_member_use
+        SlottedMultiChildRenderObjectWidgetMixin<_YaruToggleButtonSlot,
```
Jupi007, line 36: Maybe extend the `SlottedMultiChildRenderObjectWidget` class instead of still use this deprecated mixin?

Jupi007 (changes requested):

lib/src/widgets/yaru_toggle_button_layout.dart
Feichtmeier: I tried but ended up with all overwrite methods are wrong
What is this widget. Do you know where it is used?
But I can try to correctly create it later

Jupi007: Are you sure?
Because it worked fine in my PR: #767

Jupi007 (approved):
LGTM :+1:

PR #777, opened 2023-09-19
feat: togglable canvas scale

Jupi007:
Simplify the tooglables code by use canvas scale instead of manual scaling.
There's only a little look diff in active state, except that nothing changed.


Feichtmeier (approved):

Feichtmeier:
Checked out and built, no visual regression visible so I guess this is good :)

PR #778, opened 2023-09-19
chore: update example with the new paned style

Feichtmeier:
grafik

### DIFF
--- a/example/lib/example.dart
+++ b/example/lib/example.dart
@@ -67,6 +67,8 @@ class _MasterDetailPage extends StatelessWidget {
       ),
       pageBuilder: (context, index) => YaruDetailPage(
         appBar: YaruWindowTitleBar(
+          backgroundColor: Colors.transparent,
+          border: BorderSide.none,
           leading:
               Navigator.of(context).canPop() ? const YaruBackButton() : null,
           title: buildTitle(context, pageItems[index]),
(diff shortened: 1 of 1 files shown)


Jupi007 (approved):
Great!

Feichtmeier:
CC @d-loose @anasereijo eventually something to consider for the software app :) the new gtk4 apps look like this

elioqoshi:
@Feichtmeier Just to be sure, new paned style refers to the sidebar menu active item, right?

Feichtmeier:
@elioqoshi 
the sidebar now has a stronger contrast against the main part of the window
the default is still with the headerbar having the default background color, but when someone prefers to use the full paned look it can be done very easily by removing the background color and border of headerbars
this is what I meant with "paned" style :) GTK4 apps in with gnome 45 also use this style of "invisible" headerbars and the "full pane" style if it is a "master/detail" application like for example gnome control center (settings)

in this pull request here it is only changed in the example app of yaru widgets
the default still has a grey headerbar with a border, I think this is a sane default 

sorry for the long answer, needing a coffee ☕

Feichtmeier:
@elioqoshi for a better explanation I found the old PR that makes those styles possible, maybe this helps:

https://github.com/ubuntu/yaru_widgets.dart/pull/750

elioqoshi:
Ah that makes sense, thanks for the explanation!

PR #781, opened 2023-09-22
feat: add YaruTabBar

Feichtmeier:
https://github.com/ubuntu/yaru_widgets.dart/assets/15329494/4e131ca6-7836-4ce0-a541-43a7ca66d284

@Jupi007 which parameters should we expose?

### DIFF
--- a/lib/src/widgets/yaru_tab_bar.dart
+++ b/lib/src/widgets/yaru_tab_bar.dart
@@ -5,13 +5,11 @@ class YaruTabBar extends StatelessWidget {
   const YaruTabBar({
     super.key,
     required this.tabController,
-    this.icons,
-    required this.labels,
-  }) : assert(icons == null || icons.length == labels.length);
+    required this.tabs,
+  });
 
   final TabController tabController;
-  final List<Widget>? icons;
-  final List<String> labels;
+  final List<Widget> tabs;
 
   @override
   Widget build(BuildContext context) {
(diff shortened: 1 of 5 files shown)


lib/src/widgets/yaru_tab_bar.dart
```diff
@@ -0,0 +1,64 @@
...
+                    Flexible(
+                      child: Text(
+                        labels[i],
+                        overflow: TextOverflow.ellipsis,
+                      ),
+                    ),
+                  ],
+                ),
+              ),
+            ),
```
Jupi007, line 59: I think this should be extracted to a separated widget like `YaruTabBarItem`, so we can put whatever we want instead.

Jupi007 (changes requested):
Great!

I think that, for now, the only thing that we should expose is the height.

lib/src/widgets/yaru_tab_bar.dart
Feichtmeier: The icons are optional. What do you mean with anything?

Jupi007: Well, we could put any widget, instead of an icon widget and a string label.

Feichtmeier: ok so you mean instead of providing final List<Widget>? icons;
  final List<String> labels; there should be List<YaruTabBarItem> ? Or List<Widget> ?

lib/src/widgets/yaru_tab_bar.dart
```diff
@@ -0,0 +1,84 @@
+import 'package:flutter/material.dart';
+import 'package:yaru_widgets/yaru_widgets.dart';
+
+/// A pre-styled replacement for material [TabBar]
+class YaruTabBar extends StatelessWidget {
+  const YaruTabBar({
+    super.key,
+    required this.tabController,
+    required this.tabs,
+    this.height = kYaruTitleBarItemHeight + 10,
```
Jupi007, line 10: I think it would be better to allow a null value, and set the default height in a final var inside the `build` method.

Jupi007 (changes requested):
Sorry for the late reply, I have a last little request.

lib/src/widgets/yaru_tab_bar.dart
Jupi007: Yes, that's exactly what I mean 🙂

Jupi007 (approved):
LGTM 👍

PR #782, opened 2023-09-23
fix(YaruTabBar): make TabController nullable, add onTap

Feichtmeier:
"If [TabController] is not provided, then the value of [DefaultTabController.of] will be used."

~~Need to check this first if this works~~
Works, sorry, should have tested on musicpod/real app first :) 
Also added onTap, as this is part of material tabbar api

grafik

### DIFF
--- a/lib/src/widgets/yaru_tab_bar.dart
+++ b/lib/src/widgets/yaru_tab_bar.dart
@@ -5,12 +5,14 @@ import 'package:yaru_widgets/yaru_widgets.dart';
 class YaruTabBar extends StatelessWidget {
   const YaruTabBar({
     super.key,
-    required this.tabController,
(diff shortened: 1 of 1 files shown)


Feichtmeier:
Just looking at the picture, are the tabs more round here than in gtk4? Oo

Jupi007 (approved):
👍LGTM

Jupi007:
> Just looking at the picture, are the tabs more round here than in gtk4? Oo

Well, yes it looks more round 🧐
Also compared to the sidebar selected item.

Feichtmeier:
ok @Jupi007 changed to the button radius now:
grafik

Jupi007 (approved):
Much better 👍

PR #789, opened 2023-10-20
feat: add `YaruScrollViewUndershoot`

Jupi007:
Add a new `YaruScrollViewUndershoot` widget which displays an undershoot at the start and/or end of a scroll view, when the scroll offset is not at the very beginning or end. It works in both scroll axis, and you can also disable the start or end undershoot (ex: if the scroll view is close to the window border it is useless to add an undershoot).
It is also providing a `builder` named constructor which handle the `ScrollController` creation and lifecycle, so we don't need to create a statefull widget to use this.

[Capture vidéo du 2023-10-20 18-11-11.webm](https://github.com/ubuntu/yaru_widgets.dart/assets/36476595/5549203d-6676-4460-a0a6-3346f53eaa9e)

| Dark | Light |
| --- | --- |
| Capture d’écran du 2023-10-20 18-12-06 | Capture d’écran du 2023-10-20 18-11-57 |

I tried to mimic Libadwaita look:

| Dark | Light |
| --- | --- |
| Capture d’écran du 2023-10-20 18-12-35 | Capture d’écran du 2023-10-20 18-12-28 |


Feichtmeier (approved):
nice! very nice idea, cant see any mistake in the code
if you want let @d-loose double check for code

Jupi007:
@d-loose Can you quickly check this PR, please?

d-loose (approved):
LGTM :+1:

Jupi007:
Thanks @d-loose

PR #787, opened 2023-10-19
fix: overflow in left pane of `YaruMasterDetailPage`

Jupi007:
No visual change.
It looks like the problem comes from both `SliverFillRemaining` and `ListTile` which seems to not send its correct height when having a subtitle. I will investigate deeper because it looks like a bug.

@d-loose I will create a backport PR for 2.7, so you can get the fix before migrating to 3.2

Fixes #785

PR #790, opened 2023-10-23
fix: overflow in left pane of `YaruMasterDetailPage`

Jupi007:
- Revert #787 and use `IntrinsicHeight` to force real `ListTile` height computation.
  _this is a temporary fix until the real issue with `ListTile` is fixed_
- Add test variant with subtitle and spacer, to catch any future problems.

I found that the `ListTile` returns a wrong height cause to the visual density set to `VisualDensity(horizontal: -4, vertical: -4)` in `YaruMasterTile`. I will try to fix this in flutter.


d-loose (approved):
LGTM :+1:

PR #793, opened 2023-10-24
fix: final fix for left pane overflow

### LINKED ISSUE #785: Overflow in left pane of `YaruMasterDetailPage`
Using `YaruMasterTile`s with a subtitle in `YaruMasterDetailPage`'s `tileBuilder` causes an overflow in the left pane.
Discovered here: https://github.com/canonical/firmware-updater/issues/211.
Even though firmware-updater still uses yaru_widgets ^2.5.0, I confirmed this is still an issue in 3.2.0. See screencast recorded with the following change:

```diff
diff --git a/example/lib/example.dart b/example/lib/example.dart
index 7823592..c0e48c8 100644
--- a/example/lib/example.dart
+++ b/example/lib/example.dart
@@ -64,6 +64,7 @@ class _MasterDetailPage extends StatelessWidget {
       tileBuilder: (context, index, selected, availableWidth) => YaruMasterTile(
         leading: pageItems[index].iconBuilder(context, selected),
         title: buildTitle(context, pageItems[index]),
+        subtitle: Text('foo'),
       ),
       pageBuilder: (context, index) => YaruDetailPage(
         appBar: YaruWindowTitleBar(
```

[Screencast from 2023-10-12 15-42-23.webm](https://github.com/ubuntu/yaru_widgets.dart/assets/113362648/267abf64-dd48-4590-a3bd-f1556a9e909c)

Jupi007:
After investigate how `ListTile` compute its height I found that the default vertical padding was too big for the minimal visual density.
Like this the intrinsic height is correct, and the vertical alignment is also fixed.

**Before | After**

image

Fixes #785


d-loose (approved):
LGTM :+1:

PR #795, opened 2023-11-15
feat(YaruWindowControl): pimp up YaruWindowControl

Jupi007:
- Add `Windows` platform variant ;
- Fix issue with `foreground` and `background` properties (wasn't used as expected) ;
- Improve maximize/minimize animation ✨ (now done in two steps) ;
- Improve window controls example page.

https://github.com/ubuntu/yaru_widgets.dart/assets/36476595/f3cd9817-acd6-41ce-be32-f86675aedd37

The only design change (except the animation) we can notice is the minimize icon in disabled state. The little top right chevron is now toned down as expected. It is now exactly looking like its Gtk counterpart.

@Feichtmeier Should we also add a Mac Os version one day or another?


Feichtmeier (approved):
Awesome Paul thank you!


MacOS is not needed, because JP made the mac window controls somehow magically show up in the window  :D no idea why and how

Jupi007:
> Awesome Paul thank you!

Yw! 😊

> MacOS is not needed, because JP made the mac window controls somehow magically show up in the window :D no idea why and how

Okay, great 👍

Issue #799, opened 2023-11-19
Add YaruDatePicker / YaruTimePicker


Timo-Schroeder:
The Material `showDatePicker()` method opens a Dialog, which does feel a bit out of place on the desktop. Also, it does not allow the user to drag the window around.

It would be great to have a yaru Widget for this, which looks more at home on the Ubuntu desktop.

PR #824, opened 2024-01-08
feat: YaruDateTimeEntry, YaruSegmentedEntry, YaruEdgeFocusInterceptor

Jupi007:
**YaruEdgeFocusInterceptor**

Intercept focus with callbacks, so you can know if the focus is coming from a previous node or a next one.

**YaruSegmentedEntry**

An entry, split into multiple segments, which each represents a selectable editable part of the value.
This widget provides keyboard navigation, and numeric value manipulation, using keyboard tab and arrow keys.

**YaruDateTimeEntry**

A `YaruSegmentedEntry` configured to accepts and validates a datetime entered by a user.

[Capture vidéo du 2024-01-08 16-40-31.webm](https://github.com/ubuntu/yaru_widgets.dart/assets/36476595/34129bd5-9536-4730-8a31-7ce65d54e113)

**Related to:** #799

**TODO:**

- [x] Add date and time only variants ;
- [ ] Tests.

**Coming next:** Date picker overlay
image


lib/src/foundation/yaru_entry_segment.dart
```diff
@@ -0,0 +1,458 @@
...
+            _input,
+            value,
+            oldValue,
+          )
+        : value;
+    if (maxLength != null && _value != null && _value!.length > maxLength!) {
+      _value = oldValue;
+    }
+
+    if (_value != oldValue) notifyListeners();
```
Feichtmeier, line 178: Should null values always notify? (Just asking because I often need to switch these decision in various situations)

Feichtmeier (approved):
Epic work! Thank you very much
Just added a question

lib/src/foundation/yaru_entry_segment.dart
Jupi007: What do you mean exactly? I'm not sure to understand 🤔
If the value was null and becomes int or was int and becomes null it should notify, isn't it?

Feichtmeier: yes, was just asking if you want to allow to set it to null if it wasn't before :)

Jupi007: Okay, so yeah this is intended 👌🙂

PR #827, opened 2024-01-11
feat(YaruChoiceChipBar): add selectedFirst parameter

Feichtmeier:
Bildschirmfoto 2024-01-11 um 10 55 42

@Jupi007 this adds an option if you do not want the selected chips to be shown first
default stays the same in lib


Jupi007 (approved):
Great,  LGFT 👍

PR #828, opened 2024-01-16
fix(YaruChoiceChipBar): allow more parameters

Feichtmeier:
This allows to control the border color of the internal chiptheme.
This is needed because the bar includes an internal chiptheme which overwrites the one from Theme.of(context).

This internal chiptheme is needed to align the button radius to the chip radius.

Additionally this also adds options to set the background and selected background colors.


Jupi007 (approved):
LGTM 👍

PR #829, opened 2024-01-16
fix(YaruChoiceChipBar): forward parameters to buttons

Feichtmeier:
The buttons need to take the parameters too, if they would be changed.
This fixes that and also adds an optional elevation parameter


Jupi007 (approved):
LGTM 👍

PR #858, opened 2024-03-02
fix: make the icon page work in the github pages web page build

### LINKED ISSUE #857: Make the icon page work in the github pages web page build!

Jupi007:
The problem comes from the icons tree shake, so let's disable : it is not needed because we are theoretically using all the icons.
I also disabled the import of material icons, because the full file is 2Mo 😅
I replaced the only two material icons we were using in the example app, so we doesn't need to import it any more.

Fixes #857


Feichtmeier (approved):
nice, lets see if this works <3

PR #859, opened 2024-03-02
feat(YaruWindowControl)!: use material state properties for colors

Jupi007:
This PR fix the property name, and add more flexibility in the way you can customize the colors:

```dart
YaruWindowControl(
  ...
  backgroundColor:
      MaterialStateProperty.resolveWith((states) {
    if (states.contains(MaterialState.pressed)) {
      return Colors.green;
    }
    if (states.contains(MaterialState.hovered)) {
      return Colors.blue;
    }
    if (states.contains(MaterialState.disabled)) {
      return Colors.yellow;
    }
    return null;
  }),
),
```

### DIFF
--- a/lib/src/widgets/yaru_title_bar.dart
+++ b/lib/src/widgets/yaru_title_bar.dart
@@ -229,10 +229,9 @@ class YaruTitleBar extends StatelessWidget implements PreferredSizeWidget {
 
     final closeButton = YaruWindowControl(
       platform: windowControlPlatform,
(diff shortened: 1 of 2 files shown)


Feichtmeier (approved):
cool ty!
need to double check in musicpod if this hits all the weird edge cases but LGTM on the table

Jupi007:
It should be reacting in the same way as previously if you use `MaterialStatePropertyAll`.

PR #879, opened 2024-03-08
chore(example): add missing icon

Feichtmeier:
Now all icons are searched with ls -l | grep -v ^l >> icons.txt, so the list is complete. I started manually in apps when realizing that this is no fun.
Eventually the github dart package can be used to get the names automatically, if it can filter out symlinks. Otherwise it is not really a problem with the bash command.


Jupi007 (approved):
LGTM 👍

PR #880, opened 2024-03-09
feat: add YaruInfoBadge, YaruInfoBox, YaruTranslucentBorderContainer

Feichtmeier:
- Imported the design from https://github.com/canonical/ubuntu-desktop-provision/blob/main/packages/ubuntu_bootstrap/lib/widgets/info_badge.dart
- Add a translucent version of YaruBorderContainer

Using a new enum

The tip icon prbly needs a 💡 icon in YaruIcons @Jupi007 

WIP, need to add more parameters and docs

| Light | Dark |
|--------|-------|
|grafik| grafik|
|grafik|grafik|

CC @spydon


lib/src/widgets/yaru_info.dart
```diff
@@ -0,0 +1,118 @@
+import 'package:flutter/material.dart';
+import 'package:yaru/yaru.dart';
+
+enum YaruInfoType {
+  note,
```
spydon: Isn't `info` more used in this context?

lib/src/widgets/yaru_info.dart
```diff
@@ -0,0 +1,118 @@
...
+  tip,
+  important,
+  warning,
+  caution;
```
spydon: And `error` here? Caution I would expect yellow for

lib/src/widgets/yaru_info.dart
```diff
@@ -0,0 +1,118 @@
+import 'package:flutter/material.dart';
+import 'package:yaru/yaru.dart';
+
+enum YaruInfoType {
+  note,
+  tip,
```
spydon: Maybe success or something? Tip I would expect to be blue

lib/src/widgets/yaru_info.dart
Feichtmeier: I thought the same... but YaruInfoType.info reads weird :)
I got obv inspired by the github badges

> [!NOTE]  
> Highlights information that users should take into account, even when skimming.

> [!TIP]
> Optional information to help a user be more successful.

> [!IMPORTANT]  
> Crucial information necessary for users to succeed.

> [!WARNING]  
> Critical content demanding immediate user attention due to potential risks.

> [!CAUTION]
> Negative potential consequences of an action.

lib/src/widgets/yaru_info.dart
Feichtmeier: same as above, a red color could be used for other things as errors, WDYT?

lib/src/widgets/yaru_info.dart
Feichtmeier: I wouldnt limit the enum names to things we currently do
positive green things can be used for other things as a success

eventually instead of tip or success WDYT about suggestion ?

spydon: I'm not thinking that it limits it to what content the user puts in there, I'm thinking that if a user sees the four basic blocks `success`, `info`, `warning` and `error` (sometimes warning and error are bundled into one) they know what they can expect in terms of icon and color since this is a quite used design pattern. The purple is a newer thing so there doesn't seem to be any term that is "the most popular" for that, I've seen `discovery` for that one too for example, but I think `important` is fine there.

W3 uses similar terms:
https://www.w3schools.com/w3css/w3css_alerts.asp

And bootstrap too:
https://www.w3schools.com/bootstrap/bootstrap_alerts.asp

Material:
https://mui.com/material-ui/react-alert/#severity

Atlassian:
image

`danger` is a quite commonly used instead of `error` too, but the user expects red in both those cases so I don't think it matters.

Jupi007: Yes, I agree that `info`, `success`, `warning` and `error`, are the four alert types I would expect to see.

Feichtmeier: okay maybe the checkmark is a bit misleading

I have put the checkmark icon in there as a replacement until we have a light bulb :)

I can change it to information/info, but then maybe instead of 
YaruInfo.info
how about
YaruNote.info
?

danger is def. better than caution because it is stronger and leaves room for other things as errors

I am not really agreeing on the success thing.
But since you both agree I will do it 😄

Feichtmeier: I think merging the success thing with the checkmark would be bad
Any better suggestion as a replacement for the light bulb @Jupi007 ? Otherwise I can just leave it open until then, and maybe only merge the YaruTranslucentBorderContainer first?

spydon: Maybe we could call it `Alert`, instead if you don't like the repetition of words? :thinking: 

> I have put the checkmark icon in there as a replacement until we have a light bulb :)

Lightbulb I wouldn't necessarily expect on a green background, I think you are looking too hard on GitHub's design, which doesn't necessarily reflect what mobile developers would expect. :wink: 

I guess we'd want a possible override of the icon here in the long run too.

Jupi007: What about `YaruAlert`, alert is the common name on the web for that kind of component.

I would see these values:
- **info**: ℹ️ icon, blue ;
- **tip**: 💡 icon, blue ;
- **success**: ✅ icon, green ;
- **warning**: ⚠️ icon, yellow/orange ;
- **important**: "!" icon, purple ;
- **error**: "danger" icon, red.

I'm working on the light bulb icon btw:

image

spydon: > I think merging the success thing with the checkmark would be bad

How come?
Almost everyone uses a checkmark to indicate success:

image

Feichtmeier: Nice icon!

Well, neither mobile nor web should be our reference IMO. As we primarily target the desktop, even though it also works on the web great. And for the desktop our main reference is the GNOME design, even though we leave options to disagree them and put our own ideas/improvements in there. But this is eventually a bit meta/off-topic ;)

Generally it is nice to have more options. As there is already an   final Icon? icon; parameter I guess we are safe

I can leave it open until you have the light bulb done, and add the blue tip enum/box/badge after that. Then all combinations are possible but we have a set of default options we can agree on?

Feichtmeier: @spydon it's okay :) If not forced from other conditions we normally just follow democracity here and since you both agree on the checkmark I will do it. I just think it is great to have those "hey what about this idea?" boxes. I can parameterize it so all options are there but the default is what you two agree on

spydon: > What about YaruAlert, alert is the common name on the web for that kind of component.

Haha it must indicate a good naming if both of us came up with it separately. :smile: 

> I just think it is great to have those "hey what about this idea?" boxes.

If using @Jupi007's suggestion you'd have that too (not green though) :slightly_smiling_face: 

> tip: 💡 icon, blue ;

Feichtmeier: Okay so what is the plan now?
YaruInfoType/Box/Badge
YaruAlertType/Box/Badge
YaruNoteType/Box/Badge
?
and 

.info
or
.information
?

those discussions are ofc also super colored by what the native language is :)

for example "alert" in german is the same thing as "alarm" which is rather "bad", if you know what I mean

grafik
grafik

spydon: For the enum `YaruAlertType` I think is the best, and `YaruAlertBox` is is fine I think.
The badge however, I named that pretty badily to start with, maybe just `YaruBadge` would be enough there?

Feichtmeier: hm I think I have a rather strong favor of YaruInfo
As an information is rather "neutral" colored. Warning information, success information, danger/error information

I can live with YaruInfoType.information if you can too :) Better than the rather negative "alert" IMO

spydon: Yeah, I agree with you that it has a negative connotation, most others seem to use it though.
But I'm fine with info too!

Feichtmeier: cool!
I just wrap the icon Widget parameter in a

```dart
IconTheme(
    data: IconTheme.of(context).copyWith(
      size: 30,
      color: baseColor,
    ),
    child: icon ?? Icon(yaruInfoType.iconData),
  )
```

and with a final baseColor = color ?? yaruInfoType.getColor(context); optional color parameter

then I think we don't need just a special enum for the light bulb thing :) 
As I could just use the light bulb icon

Because double blue is kinda weird?

Now we even feel the colors 😆 We should add how they smell in the docs :D (little joke)

Jupi007: I don't think that `alert` sounds bad, but maybe it is because I'm used to this term :p
I'm fine with `YaruInfo` too (I prefer `Alert` though).

Feichtmeier: https://github.com/ubuntu/yaru.dart/assets/15329494/7c8cf3a4-bda6-49f0-9590-92a4d6e3421e

xD

Any further ideas? Otherwise I finish the docs :)

Jupi007: Little nitpick, but there's no "danger" badge at the end of the page.

Feichtmeier: Thanks, and you both wanted "error" instead of "danger", right?

Jupi007: Honestly I don't care, both sounds good to me :p

spydon: I'm fine with both of those options!

Jupi007 (approved):
LGTM :+1:

PR #884, opened 2024-03-10
chore(example): improve wording, placement and info

Feichtmeier:
Improved the placement of pages, wording, info and naming for people not involved into yaru yet:

- material components as the first page, as this page has the texttheme and buttons, which is important to see right away
- improved page wording to make it clear that those are pure material components but with Yaru theming
- improved text on/for components where it is not clear what type they are
- added the nice new infoboxes for coponents that have a YaruXXX replacement

Bildschirmfoto 2024-03-10 um 09 26 01
Bildschirmfoto 2024-03-10 um 09 26 30
Bildschirmfoto 2024-03-10 um 09 26 45
Bildschirmfoto 2024-03-10 um 09 26 57

### DIFF
--- a/example/lib/example.dart
+++ b/example/lib/example.dart
@@ -60,7 +60,6 @@ class _MasterDetailPage extends StatelessWidget {
   @override
   Widget build(BuildContext context) {
     return YaruMasterDetailPage(
-      initialIndex: 1,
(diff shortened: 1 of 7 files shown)


Jupi007 (approved):
LGTM 👍 

Very nice!

PR #885, opened 2024-03-10
fix(YaruMasterDetailPage): has too many nested Scaffolds

### LINKED ISSUE #728: YaruMasterDetailPage has too many nested Scaffolds
I think the main reason was to get a background, right?
There are two nested Scaffolds I don't think this is needed. Can't we replace them with Material ?

Feichtmeier:
- [X] Test on MacOS
- [X] Test on Linux
- [x] Test on Windows
- [X] Test on Web

grafik
grafik
grafik

This needs testing on all platforms, as all platforms behave differently with "draw in in the native window" before it can be considered a solution.

Note: in fact no scaffold is needed by default at all, as we only require the next child of materialapp to be a material widget. Scaffold is a very heavy widget with lots of stuff going on there. Even if the "expected" child page of one master element is a Scaffold or a YaruDetailPage, which is a Scaffold with lipstick

The normal layout with YMDP is the following
grafik

Fixes #728


Jupi007 (approved):
LGTM :+1:

spydon (approved):

PR #890, opened 2024-04-20
fix(YaruIcons): Fix emote_monkey icon broken path

Jupi007:
**Before:**

Capture d’écran du 2024-04-20 20-16-26

**After:**

Capture d’écran du 2024-04-20 20-19-17


Feichtmeier (approved):
Thanks ☺️🐒🌞

Some commas are wrong :D

Jupi007:
The icon font tool return a huge unformatted file, and I need to manually apply all the fix, sorry for the noise with all those commits '^^

Issue #814, opened 2023-12-13
Make YaruExpandable and YaruExpansionPanel optionally build children lazily 


omar-selo:
Right now it seems that even when a `YaruExpandable` or `YaruExpansionPanel` are not open, the build method is still called. I noticed this because we had a child widget that made an API request. We expected that API request to only be made when expanded, but it was made immediately. Perhaps these Yaru widgets could have a `builder` attribute that gets called when they're expanded. Of course lazy building might not suit all use cases. But it would be nice to have both options.

omar-selo:
For comparison this issue isn't present with `ExpansionPanel`. The children attribute don't actually build until they're expanded

Feichtmeier:
@Jupi007 could you eventually tackle this? I tried and failed miserably :laughing: Repaints were even worse in my branch. I checked with the nice color rebuilds option in flutter dev tools

Jupi007:
@Feichtmeier I will look at it :)

PR #892, opened 2024-05-01
fix: Make YaruExpansionPanel build lazily

Feichtmeier:
This is helps with #814 

By default all expansions are toggled but this can be disabled now if wanted, which then does not rebuild the other children.
Also the expandables are now wrapped in listview builders instead of a column.
If the scrolling needs to be disabled one can pass NeverScrollableScrollPhysics

This does not cover issues that may be in YaruExpandable

with `collapseOnExpand: true`:

https://github.com/ubuntu/yaru.dart/assets/15329494/ae11bbca-d417-4ba9-a79f-5c16e9bbb401

with `collapseOnExpand: false`:

https://github.com/ubuntu/yaru.dart/assets/15329494/d04a618b-cb8f-4b61-95f8-5c3fef48be1c

Ref #814

### DIFF
--- a/example/lib/pages/expansion_panel_page.dart
+++ b/example/lib/pages/expansion_panel_page.dart
@@ -6,32 +6,26 @@ class ExpansionPanelPage extends StatelessWidget {
 
   @override
   Widget build(BuildContext context) {
-    return YaruScrollViewUndershoot.builder(
-      builder: (context, controller) {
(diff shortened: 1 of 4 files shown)


Jupi007 (approved):
I can't test it for now, but from what I see LGTM 👍

spydon (approved):
Lgtm!

PR #909, opened 2024-08-07
chore: update flutter version to 3.24 and update dependencies

Feichtmeier:
Update to 3.24 stable

I added TODOs for the deprecated Navigator API, since we werent sure yet what to happen with YMDP.
I hope this is okay   @Jupi007 @spydon?

### DIFF
--- a/example/lib/pages/full_color_icons_page.dart
+++ b/example/lib/pages/full_color_icons_page.dart
@@ -182,6 +182,7 @@ const iconDirsToIconNames = <String, Set<String>>{
     'accessories-dictionary',
     'accessories-text-editor',
     'address-book-app',
+    'app-center',
     'applications-multimedia',
     'applications-office',
     'backups-app',
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,7 +10,7 @@ on:
   workflow_dispatch:
 
 env:
-  FLUTTER_VERSION: 3.22.0
+  FLUTTER_VERSION: 3.24.0
 
 jobs:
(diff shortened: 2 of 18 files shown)


Jupi007 (approved):
LGTM :+1: 
Thanks @Feichtmeier !

Jupi007:
> I added TODOs for the deprecated Navigator API, since we werent sure yet what to happen with YMDP.

That's a good idea, no need to spend time on something we could drop in a close future :)

Feichtmeier:
Cool! I think the goldens will be fixed tonight
I checked progress indicator and it looks like always

https://github.com/user-attachments/assets/506395bd-744b-4e68-8503-5c5c944c9bca

example/lib/pages/full_color_icons_page.dart
```diff
@@ -182,6 +182,7 @@ const iconDirsToIconNames = <String, Set<String>>{
     'accessories-dictionary',
     'accessories-text-editor',
     'address-book-app',
+    'app-center',
```
spydon, line 185: So we have a separate one called app-center? Since there is the snap-store icon too?

spydon (approved):
Lgtm, just one minor comment

example/lib/pages/full_color_icons_page.dart
Feichtmeier: I just listed all app icons from yaru. Ideal would be some http call trying to list all those icons on its own (github dart api?) but I didnt manage this so it is a static list of all icons from yaru.

PR #911, opened 2024-08-14
feat: Support GNOME accent colors

### LINKED ISSUE #908: Support GNOME accent colors
GNOME 47 will supports [accent colors](https://gitlab.gnome.org/GNOME/libadwaita/-/blob/main/src/adw-accent-color.c?ref_type=heads#L14-29), however this comes with some limitations as the number of colors and their values are limited.

We still have to define a full migration strategy, but ideally we're removing the ubuntu specific colors and use upstream ones.

We *may* tune some of the upstream colors to match the previous colors of us, but this is not yet defined (for sure we'll do it for the orange though).

So, opening this so that we can prepare a migration strategy (we'll need to listen on `xdg-setting` for the value in use), while more details need to come.

For sure yaru.dart will need to support the [same logic](https://gitlab.gnome.org/GNOME/libadwaita/-/blob/main/src/adw-accent-color.c?ref_type=heads#L159-199) to get the color to use depending on the xdg-setting value.

Feichtmeier:
TODO:

- [X] Add new variants
- [x] Detect 24.10+ systems and exclude "old" yaru variants

@3v1n0 I guess we should keep the orange from ubuntu for brand reasons?

Fixes #908


3v1n0:
Yeah, I'm also preserving the old colors in case Yaru is used, so I think you should basically using the [same logic](https://salsa.debian.org/gnome-team/libadwaita/-/commit/37de15eac23586bfc72802446144f7eb9ba81392).

Feichtmeier:
> Yeah, I'm also preserving the old colors in case Yaru is used, so I think you should basically using the [same logic](https://salsa.debian.org/gnome-team/libadwaita/-/commit/37de15eac23586bfc72802446144f7eb9ba81392).

so, if there is a blue in both yaru and adw, use the blue from yaru but with the new name? I mean the names do not matter much here because we need to map the gtk accent-color name to the flutter theme anyways

3v1n0:
Yeah, doing that for now... We're going to drop those special cases probably, but let's do this as first step

3v1n0:
Maybe outside the Ubuntu session you can just use the normal adwaita colors though.

Feichtmeier:
https://github.com/user-attachments/assets/33b842fa-d698-40f8-adc3-c1a59151b094

@3v1n0 @Jupi007 
it works, and if it is gnome but not ubuntu the adwaita version is used
I didnt even have to check for a different gtk settings string, it is still there
I just needed to map the new colors

Feichtmeier:
what warty brown though? @3v1n0 ? some sort of bonus?

spydon (approved):
Lgtm!

Jupi007:
Great!
Thanks @Feichtmeier 👍

PR #914, opened 2024-08-27
fix: use safe picks for ubuntu colors

Feichtmeier:
This string kung fu was to unreliable
I changed it to a plain mapping, just replaceAll '-dark' with ''

I tested this for ubuntu ubuntu sessions successfully 

https://github.com/user-attachments/assets/5c74fcdb-bb6d-4604-8bc9-587b9e043910

but it needs to be tested for gnome session too
@3v1n0 

Maybe I manage it in my vm then I will update this PR


spydon (approved):
Lgtm!

Jupi007 (approved):
LGTM 👍

PR #916, opened 2024-09-01
fix: rename brown to wartyBrown, matching the gtk theme name

Feichtmeier:
its called wartybrown...

grafik


Jupi007 (approved):

Feichtmeier:
Now it works :)

grafik

@3v1n0 I tested the gnome session but the settings just produced Yaru-XXX strings inside dconf-editor

So I don't know how those Adwaita theme names are
I just assumed

Adwaita
Adwaita-dark
Adwaita-pink
Adwaita-pink-dark

etc
Hope this is correct 🤞

3v1n0:
Is it possible to ensure that when used as background for a button or a checkbox wartybrown can use `#92714a` instead?

Feichtmeier:
> #92714a

and the rest a different brown?

PR #918, opened 2024-09-02
chore: renamed the snap back to yaru-widgets-example which is what's published

kenvandine:
Renamed the snap back to yaru-widgets-example which is what's published in the snap store


Feichtmeier (approved):
TY!

PR #921, opened 2024-09-17
fix: ensure better contrast for the warty version

### LINKED ISSUE #920: Ensure that when used as background for a button or a checkbox wartybrown   use `#92714a` instead
Is it possible to ensure that when used as background for a button or a checkbox wartybrown can use `#92714a` instead?

_Originally posted by @3v1n0 in https://github.com/ubuntu/yaru.dart/issues/916#issuecomment-2353365873_

Feichtmeier:
@3v1n0 only using specific versions of a color only for specific elements is ofc possible but it will probably create some side effects for the moment
So instead could we use your proposed accent color just for all elements instead?
photo_2024-09-17 21 43 21
Bildschirmfoto 2024-09-17 um 21 40 48
Bildschirmfoto 2024-09-17 um 21 40 35
Bildschirmfoto 2024-09-17 um 21 40 19

Fixes #920

### DIFF
--- a/lib/src/colors.dart
+++ b/lib/src/colors.dart
@@ -105,7 +105,8 @@ class YaruColors {
   static const Color sage = Color(0xFF657B69);
 
   /// Warty Brown
-  static const Color wartyBrown = Color(0xFFB39169);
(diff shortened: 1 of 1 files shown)


3v1n0:
Ok I think that's still better

PR #922, opened 2024-09-17
feat: follow accent-color gsettings key for 24.10

Feichtmeier:
Okay this turned out to be a bit more complicated, just listening for gtk-theme-name is not enough anymore.

For the flutter people:

- with ubuntu 24.10 comes a new gnome version which supports their own implentation of accent colors. Ubuntu supports this since 22.04 but it was a plain old "gtk-theme". Those new accent colors are exposed via portals and the value is written into a gsettings schema org.gnome.desktop.interface with the key accent-color
- thus we can not just use the ffi implemetation of "gtksettings" anymore for this (this is not gsettings, for more details ask robert)
- for the new thing we use the gsettings package (by JP and robert) which is a bit slower though since it is not ffi but there is no other way unless we create a portal service inside yaru (also possible)

For the gnomers (i.e. marco :P @3v1n0 ):
- I hope the accent-color listening is enough? now I also properly mapped those pure "blue, ..." values without any theme name inside, just as it is written inside gsettings, however I didnt manage to implement the exceptions yet for when this is gnome but not ubuntu, maybe next cycle, hope this is okay?

24.10:

https://github.com/user-attachments/assets/bed8a8cd-8f1b-435a-8ddf-c0f31212b728

24.04:

https://github.com/user-attachments/assets/4bdda218-4e18-4140-b652-092d9f52bbbc


3v1n0:
Logic seems fine, not sure on code itself :)

However, in theory applications could even just rely on dbus, using the portal value, but that would require some color translation that it's not very easy without relying on Adw itself.

Feichtmeier:
> Logic seems fine, not sure on code itself :)
> 
> However, in theory applications could even just rely on dbus, using the portal value, but that would require some color translation that it's not very easy without relying on Adw itself.

yes we can create new flutter themes on the fly with any given primary color
but currently a portal service is out of scope for me so maybe for after 24.10?

lib/src/theme_widgets/settings_service.dart
```diff
@@ -0,0 +1,104 @@
...
+    } on GSettingsUnknownKeyException catch (_) {}
+    return value;
+  }
+
+  Future<void> setValue<T>(String key, T value) async {
+    if (_values[key] == value) {
+      return;
+    }
+    _values[key] = value;
+    switch (T) {
```
spydon: Maybe use `return switch` here and then embed the cases below this switch too

spydon (approved):
Lgtm, just one minor nit

spydon (approved):
Nice, much cleaner! 🎉

lib/src/theme_widgets/settings_service.dart
```diff
@@ -0,0 +1,93 @@
...
+    T? value;
+    try {
+      _settings.get(key).then((v) {
+        value = v.toNative() as T?;
+        if (_values[key] != value) {
+          _values[key] = value;
+          notifyListeners();
+        }
+      });
+    } on GSettingsUnknownKeyException catch (_) {}
```
d-loose, line 66: Sorry, didn't have time to review this yesterday: exceptions thrown by `get(key)` won't be caught here - you either need to add a `catchError()` handler or use `async`/`await` to catch them in the `try` block. See https://dart.dev/libraries/async/futures-error-handling

Feichtmeier: yeah! just noticed this too, fixing https://github.com/ubuntu/yaru.dart/issues/923

d-loose: Awesome, thanks a lot - will have a look tomorrow!

Issue #912, opened 2024-08-16
Split button


spydon:
![image](https://github.com/user-attachments/assets/06a394ff-1614-4446-93c3-beb3a1ae9b22)

I think this type of button has been discussed before, right?
Just opening an issue here to track it.

Related: https://github.com/ubuntu/app-center/issues/1776

spydon:
Also discussed here: https://github.com/ubuntu/app-center/issues/1491

Feichtmeier:
That's fixed

PR #928, opened 2024-10-07
feat: YaruSplitButton

Feichtmeier:
Either a list of poupmenuitems needs to be provided or an onOptionButtonPressed callback

Question was if the list of options/callback should still be active if the main action is null/disabled?

Ref #912

### DIFF
--- a/example/lib/common/space.dart
+++ b/example/lib/common/space.dart
@@ -1,0 +1,20 @@
+import 'package:flutter/widgets.dart';
+
+List<Widget> space({
+  required Iterable<Widget> children,
+  double? widthGap,
+  double? heightGap,
+  int skip = 1,
+}) =>
+    children
+        .expand(
+          (item) sync* {
+            yield SizedBox(
+              width: widthGap,
+              height: heightGap,
+            );
+            yield item;
+          },
+        )
+        .skip(skip)
+        .toList();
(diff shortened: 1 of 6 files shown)


Jupi007 (approved):
LGTM :+1: 
But I think it would be better to have a more generic implementation: something like a YaruGroupedButton widget, which could be more versatile :thinking:

Feichtmeier:
> LGTM 👍 But I think it would be better to have a more generic implementation: something like a YaruGroupedButton widget, which could be more versatile 🤔

what do you mean? like the position of the dropdown?

Jupi007:
In my idea, we could add as much item as we want, with any role, just like this: https://getbootstrap.com/docs/4.0/components/button-group/#nesting

Feichtmeier:
> In my idea, we could add as much item as we want, with any role, just like this: https://getbootstrap.com/docs/4.0/components/button-group/#nesting

hm I don't know if the positioning of the menu will work then but maybe its totally possible
how about we add both: this specialized widget and that one?

Jupi007:
Yes, and this specialized widget could be a shortcut to the generic grouped button one.
This can be done in another PR :)

Feichtmeier:
ok as requested I made the menu right aligned
also removed the default width so the width is just how wide the menu wants to be
I left the width parameter so when one wants to set it its still possible

grafik
grafik
grafik

Feichtmeier:
Merged it with ref instead of fixes so we can work on it more if we want to

spydon:
Looks great! :heart_eyes:

PR #933, opened 2024-10-16
fix: Correct menu alignment for YaruSplitButton

spydon:
This fixes the issue with the menu alignment, so that the menu is always aligned with the right side of the button.

[Screencast from 2024-10-16 11-32-54.webm](https://github.com/user-attachments/assets/02f31a56-9bc9-411c-88cd-45e02e8c9c68)

@Feichtmeier  If you want I can remove the row examples from the example, I just used them for debugging.

### DIFF
--- a/example/lib/pages/split_button_page.dart
+++ b/example/lib/pages/split_button_page.dart
@@ -14,6 +14,7 @@ class _SplitButtonPageState extends State<SplitButtonPage> {
   @override
   Widget build(BuildContext context) {
     const contentWidth = 500.0;
+    const spacing = 16.0;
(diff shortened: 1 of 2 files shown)


Feichtmeier (approved):
Wow cool! This means beer for you and you need to add yourself to the page xD <3 I love myself <3 (little pun from discord :P )

I swear I had this in the beginning but for whatever reason I did swap it to center? Weird!
Thank you very much!

edit: no leave the example like you did its nice for debugging, thats what those pages are for

d-loose:
Awesome!

PR #935, opened 2024-10-17
fix(YaruPanedView): do not overwrite the divider theme

Feichtmeier:
- this should be changed inside the theme

otherwise other dividers to not match the divider inside the yaru paned view, as for example in the app center:
grafik

with this pr the paned view divider used inside the landscape layout is thinner, but most importantly it uses the definition from the theme:

Bildschirmfoto 2024-10-17 um 15 41 19


Jupi007 (approved):
LGTM 👍 
Thanks @Feichtmeier

Feichtmeier:
Cool! Ty, the goldens need to be regened afterwards

PR #943, opened 2024-10-27
fix: HighContrast YaruMasterTile, move Example theme buttons to side pane

Feichtmeier:
Apparently YaruMasterTile had some custom text color and bg due to an animated container.
Now it uses the list tile theme, this also needed a material widget behind every tile, which we should prbly have done before :P 

It was annoying to switch to the material theme page every time to use the theme buttons so I moved them out of those theme pages which needed a rework. So I also  made a change to use watch_it instead of ubuntu service and provider

https://github.com/user-attachments/assets/b54be43d-1d8a-4d1e-af3b-d4c388b834af


Feichtmeier:
The golden test fails however I checked with a color picker on macos:


https://github.com/user-attachments/assets/a18db99d-8339-4a3b-be5b-7fddc2b4ed99

Jupi007:
I think those icons should be moved to a hamburger menu, just like the text-editor app:
image

Feichtmeier:
> I think those icons should be moved to a hamburger menu, just like the text-editor app: image

I agree with you but since we do not have popovers in yaru.dart the best you can get with material is this


https://github.com/user-attachments/assets/fdd2fbbd-ccd1-478b-9f85-403fef94ed5d

since this PR fixes an issue with YaruMasterTile, can we do this in a follow up?

Jupi007 (approved):
LGTM :+1:

PR #958, opened 2024-11-05
feat: new indeterminated progress design

Jupi007:
[Capture vidéo du 2024-11-06 00-28-34.webm](https://github.com/user-attachments/assets/ea2a619d-9d93-4d22-b5c5-03237ed0fc43)

Note: the determinate linear progress is too big, it's a side effect I need to fix but don't take attention on that. I also need to add an initial animation to the linear undetermined animation

@Feichtmeier You can give me feedback by testing this branch :)


Feichtmeier:
it looks really good! awesome transition!!!

I wonder if it would be doable to enable/disable the dots showing up? maybe without the dots/discs it would be a good default because our theme is also used in the ubuntu flavours and it somehow "hardcodes" the ubuntu logo into it? Maybe this is overthinking I dont know

Jupi007:
Argh, I didn't thought about this.... :/
Do we provide a way to detect ubuntu flavor in yaru? I don't remember...

Feichtmeier:
> Argh, I didn't thought about this.... :/ Do we provide a way to detect ubuntu flavor in yaru? I don't remember...

yes with the platform plugin I use in the inherited theme, but instead I would just make the dots optional and default to the look we have here 
https://github.com/ubuntu/yaru/pull/4160

Jupi007:
@Feichtmeier I think it is a bad idea to change the design depending on the platform. Technically it would have been nice, but it would have caused sizing problems: the dots make the widget a bit bigger.

I did a similar, but simpler, design without dots:

https://github.com/user-attachments/assets/cd40ebea-56cf-4484-8826-add402b60fce

This is also better because the GTK version doesn't have dots, so it would have looked weird to have dots in Flutter.

Feichtmeier:
is it rdy to merge? or do you want to wait until this is in gtk ?

Jupi007:
I need to cleanup the code a little bit, then we can merge :)

Feichtmeier:
test

Jupi007:
Okay, now it works. @Feichtmeier you can review and merge.
The golden tests only fail cause to the windows controls PR.

Feichtmeier (approved):
Thank you very much, LGTM!

PR #960, opened 2024-11-06
docs : Added snap insallation proccess

MrSIHAB:
Snap is more simple and better way to install flutter. It's a good solution for beginners to advace user.


Feichtmeier:
hi! yes this would be our desired way too, however due to our apps modifying the myapplication.cc with native libraries this does not work well, this is why we do not recommend it for desktop installations (currently)

PR #962, opened 2024-11-14
fix(Chips): chip height altered by wrong label style, padding at the right

Feichtmeier:
The wrong label style made the chips label be placed not in the center 
also in the choice chip bar the padding was always at the end, which is wrong

Before:

https://github.com/user-attachments/assets/9655e8f4-cd34-494e-b25a-c508600ba8a7

After:

https://github.com/user-attachments/assets/96732cca-964d-41df-893d-8248f7bbf301


Jupi007 (approved):
LGTM 👍

PR #967, opened 2024-12-03
fix: Square Windows title controls

ashuntu:
When using the YaruWindowTitleBar on Windows, the title bar buttons (minimize/maximimze/close) appear as rectangles pushed to the top right corner of the window. As seen here for the [Ubuntu Pro for WSL](https://github.com/canonical/ubuntu-pro-for-wsl/tree/main/gui/packages/ubuntupro) app:

image

However, a lot of Windows apps actually have the buttons fill the height of the title bar, as seen here in Windows terminal:

image

For our app we also would prefer these full-height controls as opposed to the rectangular and half(ish) height controls. There currently is no way to alter button sizes, only the padding of the buttons. This PR changes the buttons to be square, with the same width/height as the title bar height. In addition, this change still allows the former behavior by default, since to have full height controls the caller still needs to pass `EdgeInsets.zero` to `buttonPadding` of the `YaruWindowTitleBar` like so:

```dart
YaruWindowTitleBar(
  title: Text(lang.appTitle),
  buttonPadding: EdgeInsets.zero,
)
```

This PR also updated a test to accommodate this change, as well as the golden images for Windows title bar or control related images.

Here is a comparison of behavior:
| | Default | Passing zero buttonPadding |
|----------|-------------------|----------------------------|
| This PR | image | image |
| Existing | image | image |


Feichtmeier:
@ashuntu hi thanks 👍 
Is 46 the size that is always the correct height? Otherwise we could make it expand?

ashuntu:
> Is 46 the size that is always the correct height? Otherwise we could make it expand?

46 is the height of the title bar (`kYaruTitleBarHeight`). I could just make `kYaruWindowsWindowControlSize = kYaruTitleBarHeight`?

Feichtmeier (approved):
I think it's fine. When you find windows apps with different sizes we can adjust it

Feichtmeier:
The tests will be fixed tonight with the nightly

Jupi007:
Great, thanks! @ashuntu 
Glad to see you use our windows window controls :D

PR #970, opened 2024-12-18
feat(example): monospace text style in `_CodeDialog`

andrew-ares:
Use monospaced style in `_CodeDialog` to improve readability.

yaru_example_2024_12_18@2x


Jupi007:
Great 👍
But maybe should we import the ubuntu mono font into this project directly.
Byw this could be helpful for the datetime picker.
Wdyt @Feichtmeier ?

andrew-ares:
Please take another look

Feichtmeier:
I think the font adding in the yaru pubspec is enough, the example should then use the font by using the yaru package from path

Jupi007:
Yes I think you're right @Feichtmeier

Feichtmeier (approved):

PR #977, opened 2025-03-14
feat: reduce YaruInfoBox titleTextStyle default height

matthew-hagemann:
Hi @Feichtmeier :slightly_smiling_face: 

While making use of the `YaruInfoBox`, @juanruitina and I found the default value for `height` for the `titleTextStyle` was a bit high and lead to the text wrapping leaving more space between the two lines of the wrapped title text than between the title and the subtitle:
Screenshot From 2025-03-13 12-59-14

After some playing around with values, 1.3 seems to get better results:

Screenshot From 2025-03-13 17-46-50

We currently are just overriding the style on the title, but thought it would be valuable to upstream as a default.


Feichtmeier (approved):

Feichtmeier:
Thank you!

PR #980, opened 2025-03-19
fix: YaruExpandable: handle tap on header background

jpnurmi:
A proposal to allow expanding and collapsing `YaruExpandable` and `YaruExpansionPanel` by tapping anywhere on the header, not just the header child and the icon button.

The interactive area visualized before vs. after:

| Before | After |
|--------|-------|
| Screenshot From 2025-03-19 15-17-19 | Screenshot From 2025-03-19 15-14-08 |

This small behavioral change makes a big difference, especially on the accessibility page in the desktop installer (before, clicking where the cursor is would do nothing):

image

### DIFF
--- a/lib/src/widgets/yaru_expandable.dart
+++ b/lib/src/widgets/yaru_expandable.dart
@@ -95,9 +95,7 @@ class _YaruExpandableState extends State<YaruExpandable> {
       ),
     );
 
-    final header = Flexible(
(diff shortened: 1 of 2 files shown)


Jupi007 (approved):
LGTM, this is the behaviour I expect when clicking in the empty space 👍

Feichtmeier (approved):
Am I dreaming? 
WB! 
And thanks for your PR :)

PR #982, opened 2025-04-05
feat(YaruExpansionPanel): add optional initialIsExpanded property

Feichtmeier:
https://github.com/user-attachments/assets/efa483eb-8e6d-4937-adde-e61e18ed5d2b

### DIFF
--- a/lib/src/widgets/yaru_expansion_panel.dart
+++ b/lib/src/widgets/yaru_expansion_panel.dart
@@ -14,6 +14,7 @@ class YaruExpansionPanel extends StatefulWidget {
         const BorderRadius.all(Radius.circular(kYaruContainerRadius)),
     this.border,
     required this.headers,
+    this.isInitiallyExpanded,
     this.width,
     this.height,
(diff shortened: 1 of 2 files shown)


lib/src/widgets/yaru_expansion_panel.dart
```diff
@@ -14,6 +14,7 @@ class YaruExpansionPanel extends StatefulWidget {
         const BorderRadius.all(Radius.circular(kYaruContainerRadius)),
     this.border,
     required this.headers,
+    this.initialIsExpanded,
```
spydon: I would name it `initiallyExpanded` or `isInitiallyExpanded`

spydon (approved):

soumyaDghosh:
Should all the other expansion panels be closed if another one is expanded?

Feichtmeier:
> Should all the other expansion panels be closed if another one is expanded?

thats a property you can generaly set  `collapseOnExpand`which is true by default

PR #983, opened 2025-04-12
feat: adapt border radii to gnome design

Feichtmeier:
@anasereijo @juanruitina @Jupi007 @Muqtxdir 

the border radii difference between gtk4apps/gnome shell  and yaru flutter apps became a little bit high imo
WDYT are these radii okayfor you?

@Muqtxdir are these radii correct?

Bildschirmfoto vom 2025-04-12 14-38-24
Bildschirmfoto vom 2025-04-12 14-38-31
Bildschirmfoto vom 2025-04-12 14-38-41
Bildschirmfoto vom 2025-04-12 14-47-03
Bildschirmfoto vom 2025-04-12 14-47-16
Bildschirmfoto vom 2025-04-12 14-47-32
Bildschirmfoto vom 2025-04-12 14-47-53


Jupi007:
To be very honest, I really prefer the less rounded style, especially for checkboxes '^^
Is it really important to be that sync with Gnome on that point?
I mean, now with flutter we have the opportunity to express our design tastes without being annoyed with Gnome compatibility like with our good old yaru Gtk3/4 theme :D
I hear the consistency argument, but personally it doesn't hurt me to have slight differences like this between flutter apps and Gtk ones. Besides, it reinforces the Ubuntu identity.

juanruitina:
I think it's acceptable to deviate if we have a good, functional reason. For instance, checkboxes should be perceived as squared so they can be differentiated from radio buttons, and I would be OK with not changing the radii in that specific case.

But for all other cases, I'm with @Feichtmeier in that we should aim for consistency. Whether we like it or not, a lot of the Ubuntu experience is currently defined by GNOME apps. If @Muqtxdir can confirm the radii values match those in Yaru GTK/Adwaita, then I'm OK with these changes.

juanruitina:
I think it's acceptable to deviate if we have a good, functional reason. For instance, checkboxes should be perceived as squared so they can be differentiated from radio buttons, I think it would be OK to keep the previous radii in that case.

But for all other cases, I'm with @Feichtmeier in that we should aim for consistency. In the end, much of the Ubuntu experience is currently defined by GNOME apps; if we wanted the visual language for Ubuntu to deviate further from what we are inheriting from Adwaita, I'd advocate for making that happen on both Yaru GTK and Flutter, not on Flutter alone.

I'm OK with these changes, once @Muqtxdir confirms the radii values match those in Yaru GTK.

Jupi007:
> I'd advocate for making that happen on both Yaru GTK and Flutter, not on Flutter alone.

I fully agree, but sadly we doesn't have any control any more on gtk theming with libadwaita apps. 

> For instance, checkboxes should be perceived as squared so they can be differentiated from radio buttons, I think it would be OK to keep the previous radii in that case.

The checkboxes are the only place where I have a strong opinion about keeping the previous radius.
In the other places, while I prefer the current look, I agree that it would create inconsistencies.

Btw, if the radius is increased in the theme, should we also update the icon theme?

Muqtxdir (approved):
Apologies, for the delayed response - yes, this matches the radii from gnome ui

Feichtmeier:
@juanruitina @Jupi007 @Muqtxdir so the general consensus here is now to go with those border-radii except check boxes?

What radius do you want for those? JUst 4 as before?
grafik
grafik

Jupi007:
The previous radius for checkboxes, for anything else the new radius is fine 👍

PR #986, opened 2025-05-04
chore: use old check radius

Feichtmeier:
@Jupi007 sorry... I just asked and didnt commit 🤦🏻 
too much computer switching


Jupi007 (approved):
Haha, no problem I understand 😄

Issue #979, opened 2025-03-18
Add focus ring to all components for improved accessibility


juanruitina:
To be compliant with WCAG Level AA, focused elements should be clearly different from their unfocused state. We seem to have inherited some insufficient focus states from Material. This was already flagged up for buttons in https://github.com/ubuntu/yaru.dart/issues/849. I suggest a general approach to the focus indicators in Yaru components: using solid 2px-wide focus rings.

I'm fully aware implementing this for all components would be a big piece of work and take time, but it would be great if we could agree on the approach.

## Rationale

[Success Criterion 2.4.13 Focus Appearance](https://www.w3.org/WAI/WCAG22/Understanding/focus-appearance.html) provides some guidance on good focus indicators:

> When the keyboard [focus indicator](https://www.w3.org/WAI/WCAG22/Understanding/focus-appearance.html#dfn-focus-indicator) is visible, an area of the focus indicator meets all the following:
> 
> - is at least as large as the area of a 2 [CSS pixel](https://www.w3.org/WAI/WCAG22/Understanding/focus-appearance.html#dfn-css-pixel) thick [perimeter](https://www.w3.org/WAI/WCAG22/Understanding/focus-appearance.html#dfn-perimeter) of the unfocused component or sub-component, and
> - has a contrast ratio of at least 3:1 between the same pixels in the focused and unfocused states.

Changing background on focus, as most components in Yaru.dart do, is generally insufficient to achieve the 3:1 contrast requirement. On top of that, we rely on background changes for hover.

The Yaru GTK theme does use focus rings. These rings use transparency (code [here](https://github.com/ubuntu/yaru/blob/7db6f52042fa3c5bb0bb7f843848c51569e21e94/gtk/upstream/gtk-4.0/Default/_colors.scss#L74-L103) and [here](https://github.com/ubuntu/yaru/blob/7db6f52042fa3c5bb0bb7f843848c51569e21e94/gtk/src/default/gtk-3.0/_tweaks.scss#L161-L204)): this makes the rings lighter visually and perhaps more appealing for some users, but also result in insufficient contrast when high contrast mode is off, <3:1 (rings are slightly more opaque when high contrast mode is on).

[...]

PR #990, opened 2025-05-08
feat: Add optional focus ring to YaruMasterTile

ashuntu:
Adds the ability to provide a `Decoration` based on focus state to `YaruMasterTile`s. This can be used to satisfy focus ring requirements. The default usage/style remains unchanged unless the new `decoration` or `focusDecoration` fields are provided by the caller.

Yaru could implement a default focus ring, but I figured that would be undesirable so I have left it up to the caller to provide a focus decoration for now.

I'd imagine other `ListTile`s in this repo could utilize a same (if not identical) widget, but I figured this would be a good place to start discussion on implementation details. I can open PRs in the future for further `ListTile` work if this implementation is sufficient.

Relevant issue: https://github.com/ubuntu/yaru.dart/issues/979 (does not close)

Example using the [App Center](https://github.com/ubuntu/app-center) (left nav):

```dart
// Decoration provided to focusDecoration
BoxDecoration(
  border: Border.all(
    color: theme.primaryColor,
    width: 2,
  ),
  borderRadius:
      const BorderRadius.all(Radius.circular(kYaruButtonRadius)),
  color: backgroundColor,
);
```

image

### DIFF
--- a/lib/src/widgets/master_detail/yaru_master_tile.dart
+++ b/lib/src/widgets/master_detail/yaru_master_tile.dart
@@ -1,5 +1,5 @@
 import 'package:flutter/material.dart';
-import 'package:yaru/constants.dart';
+import 'package:yaru/yaru.dart';
 
 const double _kScrollbarThickness = 8.0;
 const double _kScrollbarMargin = 2.0;
@@ -17,6 +17,8 @@ class YaruMasterTile extends StatelessWidget {
     this.subtitle,
     this.trailing,
     this.onTap,
+    this.decoration,
+    this.focusDecoration,
   });
 
   /// See [ListTile.selected].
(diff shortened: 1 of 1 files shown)


Feichtmeier:
HI @ashuntu thanks for your help

Generally I would prefer if this could be fixed inside the flutter/flutter repo, but as a band-aid fix for canonical products I dont have any veto against this.

Basically YaruMasterTile is just a ListTile, specified to sit inside a YaruMasterDetailPage. So the prefered way of styling would be to style it via theme. Since this here involces more than just the looks, could we eventually create a more generic widget/container that can be re-used for more widgets?

ashuntu:
> Generally I would prefer if this could be fixed inside the flutter/flutter repo

Yeah I can look into doing that as well, though obviously it may take more time and effort to get in our apps.

> could we eventually create a more generic widget/container that can be re-used for more widgets?

Definitely. I started with the YaruMasterTile since it was a higher priority issue for the App Center, but if this is an implementation that seems good I can make it more generic for other ListTiles used in Yaru as well.

Feichtmeier:
> > Generally I would prefer if this could be fixed inside the flutter/flutter repo
> 
> Yeah I can look into doing that as well, though obviously it may take more time and effort to get in our apps.
> 
> > could we eventually create a more generic widget/container that can be re-used for more widgets?
> 
> Definitely. I started with the YaruMasterTile since it was a higher priority issue for the App Center, but if this is an implementation that seems good I can make it more generic for other ListTiles used in Yaru as well.

alternatively after a review, we could get this one in, and in the next PR you could make it re-useable for other yaru widgets you and we use in the apps?

ashuntu:
> alternatively after a review, we could get this one in, and in the next PR you could make it re-useable for other yaru widgets you and we use in the apps?

Yep that sounds good!

Jupi007 (approved):
It looks good to me :+1:
But yeah, I fully agree with @Feichtmeier this would require a more generic/reusable solution.

PR #1000, opened 2025-06-09
feat(a11y): Add field for a semantic label for all icon buttons

ashuntu:
Adds the option to specify a semantic label for all icon buttons (hopefully I got them all, let me know if I missed some :)). As stated in https://github.com/ubuntu/yaru.dart/pull/999, this is useful since it allows `yaru.dart` users to retain the Yaru style and icons while providing more accessible labels to buttons.

This is a non-breaking change, it only adds an optional field to widgets that use icon buttons.

### DIFF
--- a/lib/src/widgets/yaru_search_field.dart
+++ b/lib/src/widgets/yaru_search_field.dart
@@ -30,6 +30,7 @@ class YaruSearchField extends StatefulWidget {
     this.controller,
     this.focusNode,
     this.clearIcon,
+    this.clearIconSemanticLabel,
   });
 
   /// Optional [String] forwarded to the internal [TextEditingController]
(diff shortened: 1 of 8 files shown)


lib/src/widgets/yaru_search_field.dart
```diff
@@ -396,13 +406,15 @@ class YaruSearchButton extends StatelessWidget {
...
                 color: theme.colorScheme.onSurface,
+                semanticLabel: semanticLabel,
               ),
           icon:
               icon ??
               Icon(
                 YaruIcons.search,
                 size: kYaruIconSize - 4,
                 color: theme.colorScheme.onSurface,
+                semanticLabel: semanticLabel,
```
ashuntu: We could add another field to be able to specify a unique separate label for the `selectedIcon`?

Feichtmeier: sure good idea, even though I am not really sure if we should continue on the searchbar and maybe better let people implement their own searchbars for example with the material widget (which maybe needs a bit theming treatment) but for now sure add it !

Feichtmeier (approved):

Feichtmeier (approved):

PR #1006, opened 2025-07-02
fix(a11y): Fix semantics being discarded for each Navigator

### LINKED ISSUE #997: YaruMasterDetailPage navigation isn't screen readable
In the example provided by this repository, as well as in the App Center, the left navigation of `YaruMasterDetailPage`s doesn't seem to be screen readable. I've tried various `Semantics` shenanigans in the App Center and I have been unable to get it to work. I've also briefly attempted to get the example project to read the navigation and was unsuccessful.

I think it would be sensible for the example project to by default read the text inside the list tile navigation (their titles).

ashuntu:
The `YaruMasterDetailPage` has an issue where the left navigation buttons are invisible to screen readers. For example, if you wrap the `ExampleHome` app in `SemanticsDebugger` you can see the button labels are not present:

image

This PR uses [this suggestion](https://github.com/flutter/flutter/issues/55758#issuecomment-620920637) from a Flutter issue describing this problem as a problem with the `Navigator` and how it builds routes. Now, the semantics look like this:

image

I've tested this with both the example app provided by this library and the [App Center](https://github.com/ubuntu/app-center), and it fixes both apps while retaining the same functionality.

Fixes https://github.com/ubuntu/yaru.dart/issues/997

### DIFF
--- a/lib/src/widgets/master_detail/yaru_landscape_layout.dart
+++ b/lib/src/widgets/master_detail/yaru_landscape_layout.dart
@@ -133,25 +133,29 @@ class _YaruLandscapeLayoutState extends State<YaruLandscapeLayout> {
         context,
       ).copyWith(pageTransitionsTheme: theme.landscapeTransitions),
       child: ScaffoldMessenger(
-        child: Navigator(
-          key: widget.navigatorKey,
-          initialRoute: widget.initialRoute,
-          onGenerateRoute: widget.onGenerateRoute,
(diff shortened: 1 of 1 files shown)


cnihelton (approved):
This kind of change is better viewed with the 'Hide whitespaces' setting in GH, otherwise it seems bigger than it really is. 

Simple and effective fix. 👍🏽

Feichtmeier:
Thanks!

PR #1009, opened 2025-07-25
fix: availableWidth should be updated when pane resized

adil192:
**Bug**:
When resizing the side pane of `YaruMasterDetailPage`, the `tileBuilder` is not rebuilt with the new `availableWidth`.

**Usage scenario**:
I wanted to remove the label when the pane gets small, so there's no overflow, like this:
image
image
```dart
  leading: tabs[index].iconBuilder(context, selected),
  title: (availableWidth > 120) ? Text(tabs[index].title) : null,
),
```
The availableWidth in the tileBuilder was not updated when the pane is resized. But manually triggering a hot reload calls the tileBuilder with the new width.

**Fix**:
I wrapped the pane size in a ValueNotifier, and added a ValueListenableBuilder to automatically rebuild the tile when the pane size changes.

***

**Issue template**:

> For a bug fix, please target the `release` branch, else target `main`.

The issue template says to target the `release` branch but it seems out of date. Let me know if this PR needs changing from `main`

### DIFF
--- a/lib/src/widgets/master_detail/yaru_landscape_layout.dart
+++ b/lib/src/widgets/master_detail/yaru_landscape_layout.dart
@@ -41,7 +41,7 @@ class YaruLandscapeLayout extends StatefulWidget {
 
 class _YaruLandscapeLayoutState extends State<YaruLandscapeLayout> {
   late int _selectedIndex;
-  double? _paneWidth;
+  late final _paneWidth = ValueNotifier<double?>(0.0);
 
   @override
   void initState() {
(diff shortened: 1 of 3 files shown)


Feichtmeier:
Thanks 🙏 @adil192 !

Nice idea and since you also dispose it I don't think there's a downside to this. 
Nevertheless I would like @Jupi007 to have a look at this too

Jupi007 (changes requested):
Hi @adil192 and thanks for your PR!
I think this is a wider problem. I explain myself: `YaruLandscapeLayout` relies on `YaruPanedView` for the pane resizing so this issue should be fixed there.
IMO the best solution would to add `paneBuilder` and `pageBuilder` properties to `YaruPanedView` and also a `YaruPanedView.builder` constructor.

PR #1011, opened 2025-07-26
feat: add YaruPanedView.builder constructor

adil192:
**Bug**:
When resizing the side pane of `YaruMasterDetailPage`, the `tileBuilder` is not rebuilt with the new `availableWidth`.

**Fix**:
The `paneBuilder` and `pageBuilder` are now passed directly to a `YaruPanedView.builder()` constructor, which automatically rebuilds since it uses a LayoutBuilder.

**Usage scenario**:
I wanted to remove the label when the pane gets small, so there's no overflow, like this:
image
image
```dart
  leading: tabs[index].iconBuilder(context, selected),
  title: (availableWidth > 120) ? Text(tabs[index].title) : null,
),
```
The availableWidth in the tileBuilder was not updated when the pane is resized. But manually triggering a hot reload calls the tileBuilder with the new width.

**More information**:
Replaces #1009 to address @Jupi007's [comment](https://github.com/ubuntu/yaru.dart/pull/1009#pullrequestreview-3057962458). Test files are identical to the previous PR. (I've also tested manually with my app)

### DIFF
--- a/lib/src/widgets/master_detail/yaru_landscape_layout.dart
+++ b/lib/src/widgets/master_detail/yaru_landscape_layout.dart
@@ -41,7 +41,6 @@ class YaruLandscapeLayout extends StatefulWidget {
 
 class _YaruLandscapeLayoutState extends State<YaruLandscapeLayout> {
   late int _selectedIndex;
-  double? _paneWidth;
 
   @override
   void initState() {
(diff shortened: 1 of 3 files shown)


Jupi007 (approved):
LGTM 👍 
Thanks a lot @adil192

PR #1013, opened 2025-07-27
fix: use scaffold background color instead of hardcoded color and kYaruWindowRadius for titlebars

### LINKED ISSUE #1008: YaruWindowTitleBar does not use custom ThemeData colors for focused background
Hi!

When using a custom dark theme (not based on yaru's default light/dark surface colors), the YaruWindowTitleBar does not pick up my theme's color for the focused background. Instead, it always uses the hardcoded YaruColors.titleBarDark or YaruColors.titleBarLight.

See here: https://github.com/ubuntu/yaru.dart/blob/6d7a650d59beb2e8b8f7fc72c8ef677ffe451a0b/lib/src/widgets/yaru_title_bar.dart#L177-L182

Is this intended? If not, could the title bar be adjusted to use the current ThemeData by default, so it works better with custom color themes?

I also noticed that when setting the backgroundColor property on YaruWindowTitleBar, it does override the internal focus logic and applies the same color for both the focused and the unfocused states. It would be nice if there were a way to provide different colors for both states, or for the widget to use the theme’s color scheme by default. 

This would make it so much easier to integrate with custom themes.

Thanks! :)

Feichtmeier:
@juanruitina

the titlebar colors was hardcoded. i wondered what would be the best theme color as a default (it can be manually overwritten). I thought since you guys go with this look in the installer this might be a good choice?

Bildschirmfoto 2025-07-27 um 11 43 55

Fixes #1008

### DIFF
--- a/example/lib/example.dart
+++ b/example/lib/example.dart
@@ -51,7 +51,6 @@ class _MasterDetailPage extends StatelessWidget {
       ),
       pageBuilder: (context, index) => YaruDetailPage(
         appBar: YaruWindowTitleBar(
-          backgroundColor: Colors.transparent,
           border: BorderSide.none,
           leading: Navigator.of(context).canPop()
               ? const YaruBackButton()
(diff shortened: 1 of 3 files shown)


Jupi007 (approved):
LGTM 👍

juanruitina:
Sorry for the late reply. This makes sense, thank you!

PR #1017, opened 2025-08-06
fix(a11y): Accessible switch colors and borders

ashuntu:
This PR attempts to make switch colors and borders more accessible by adding borders in light mode and making on/off shapes more contrast-y.

Note: these changes still aren't quite matching Ubuntu GNOME's colors since we don't use the theme color in high contrast mode, but that change would be much more systemic to make.

Examples (note these mainly just affect light mode):

**This PR**
| | Dark | Light |
|-|------|-------|
| Normal | image | image |
| High Contrast | image | image |

**Existing**
| | Dark | Light |
|-|------|-------|
| Normal | image | image |
| High Contrast | image | image |

**Ubuntu GNOME**
| Dark | Light |
|------|-------|
| image | image |


ashuntu:
cc @juanruitina if you have any input on this change.

juanruitina:
Yes, this ensures every element in the Yaru.dart switch has sufficient contrast in high contrast mode (>3:1) either with borders or shape colour.

I also think removing borders from the disabled state helps a lot with differentiating it from the rest. I wonder though why the colours of thumb and shape in the disabled state are inverted in the light theme (they aren't on dark theme). In fact, I think the only style change for denoting the disabled state should be to set the whole component to 50% opacity, preserving all colours, same as GTK does. This also has the benefit of working perfectly both on light and dark theme. Could we try that?

Keep me posted on any progress with honouring the On/Off Shapes accessibility setting on Flutter.

Jupi007 (approved):
LGTM, thanks @ashuntu 👍

PR #1023, opened 2025-08-20
fix(a11y): Increase page indicator contrast

ashuntu:
Contrast should be at least 3:1 for the inactive dots, this brings the contrast in light mode just above 3:1, and contrast in dark mode is just above 6:1.

| | Dark | Light |
|-|------|-------|
| This PR | image | image |
| Current | image | image |

---

UDENG-7682


ashuntu:
cc @juanruitina for visibility

juanruitina:
Looks good to me. It would be OK to increase the opacity on high contrast mode only, but I don't see the reason why not to do it at all times as you seem to be doing.

Jupi007 (approved):
LGTM 👍 
Thanks @ashuntu

PR #1026, opened 2025-08-25
feat(YaruExpandable): add `usePadding` flag

### LINKED ISSUE #1024: YaruExpandable uses an unnecessary Column
`YaruExandable` uses a `Column` to add padding with a `SizedBox` ([ref](https://github.com/ubuntu/yaru.dart/blob/main/lib/src/widgets/yaru_expandable.dart#L154-L161)), but this `Column` could be replaced with `Padding` and insets on the top side.

For context, this came up during @CarlosNihelton's PR https://github.com/canonical/ubuntu-pro-for-wsl/pull/1285, in which we had to add the child within `Row(Expanded(child))` to get the child to not be center-aligned (any content less than full width is centered by default because of the `Column`).

Either way works fine, but sometimes when you're using this widget from an application, you do not expect your child to be put into a `Column`, which can lead to issues if the child doesn't have a well defined size or if the child isn't full width. This would be a breaking chance technically, but thought I'd open it for discussion.

Jupi007:
If true, a `Padding` will be used instead of a `Column` around the child widget to add a gap between `header` and `child`. This avoids unexpected behaviour with `Column`, especially with a child that is not well sized.

In the next major release, we can use `Padding` as the default behaviour, and remove this flag.

Fixes #1024

### DIFF
--- a/example/lib/pages/expandable_page.dart
+++ b/example/lib/pages/expandable_page.dart
@@ -16,6 +16,7 @@ class ExpandablePage extends StatelessWidget {
           padding: const EdgeInsets.all(kYaruPagePadding),
           children: const [
             YaruExpandable(
+              usePadding: true,
               header: Text(
                 'Lorem ipsum dolor sit amet',
                 style: TextStyle(fontWeight: FontWeight.bold),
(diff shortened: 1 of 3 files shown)


ashuntu (approved):
Thanks for this

PR #1035, opened 2025-09-29
fix: Spacing tweaks

ashuntu:
This is a follow-up to some discussion resulting from https://github.com/ubuntu/app-center/pull/1976#issuecomment-3323175032.

This does a couple things:
- Makes the scrollbar thinner by default
- Scrollbar thickens on hover/drag/focus
- Reduces the margin around `YaruMasterTile`s (and removes its dependence on scrollbar thickness)
- Exposes `YaruMasterTile`'s `Padding` as an optional field

cc @juanruitina

| Before | After |
|--------|-------|
| image | image |

And here is what hovering the scrollbar looks like now:

image

Finally, for reference, here is the Ubuntu Settings app and how it looks with tiles/scrollbars:
image

---

UDENG-8146


Jupi007 (approved):
Great, LGTM 👍 
The scrollbar looks much better like this!

Issue #995, opened 2025-05-28
Simplify default font sizes


juanruitina:
I've noticed over the last few months how we use many different font sizes across components. As a result, it is frequent that in the same screen we have different font sizes that do not respond to hierarchy (and sometimes convey a _wrong_ sense of hierarchy), also making them visually unappealing.

See for instance this design @d-loose is currently building based on a design on mine, how the default font size of the text in the input field and the checkbox label is different to that of the rest of the page, for no good reason:

![Image](https://github.com/user-attachments/assets/c0680f0f-750f-457d-b6f9-818712f3e731)

And this is what it looks like with overrides so font sizes make sense:

![Image](https://github.com/user-attachments/assets/78bd9e36-50ce-4ebf-a71d-bc0b512704d5)

(Another example was [this dialog](https://github.com/canonical/firmware-updater/pull/326))

I quite agree with GNOME's approach for instance in the Settings app, basically using the equivalent to bodyMedium for almost everything (even headings!). See here how everything on the page has the same font size except the lighter gray descriptions below some options (clearly secondary text, makes sense hierarchically). It even allows for slightly higher information density:

![Image](https://github.com/user-attachments/assets/2da5c5de-fe36-4cae-9471-5530b5c5682f)

I think that bodyMedium should be the default font size for:

- Labels for radio buttons, checkboxes, switches
- Text in all input fields (not labels)
- Items in the sidebar
- Titles in yaruInfo panels and tiles/rows (hierarchy can be conveyed by either use of bold for title or lighter, smaller text for the description)

I think it would be OK to keep bigger font sizes for headings, given that the bigger size indeed conveys hierarchy. I also think there is a case for developers intentionally overriding font sizes in some circumstances. But I think it would be good to have simpler defaults.

(I did notice that bodyMedium is slightly smaller than GNOME's default font size, but that's a separate problem we can look into, probably related to units)

Feichtmeier:
I agree with the proposal!

My only question is where we should do this. In the widgets (if not existent create them) or the text theme?
CC @d-loose @Jupi007 @ashuntu @spydon @jpnurmi

ashuntu:
From briefly looking at some of these widgets, it seems like most of these can be solved by overriding the ThemeData for specific widget types, like `inputDecorationTheme` or `listTitleTheme`.

I know for the Installer and App Center we use manual size overrides somewhat often so it would be nice to rely more heavily on Yaru theming to be consistent.

[...]

PR #996, opened 2025-06-02
feat: use newer Ubuntu Sans `variable` fonts

Muqtxdir:
#### description:

- use the newer variable Ubuntu Sans font-family

PR #1036, opened 2025-09-30
fix: Tweak default font size to match Yaru GTK and remove letter spacing

juanruitina:
Some fixes for better consistency with GTK and better typography.

## GTK points vs Flutter pixels

GTK uses points (pt) as units, whereas Flutter uses dp (~pixels). Default text in the Yaru GTK theme is 11pt, which translate into 14.66...dp as opposed to 14dp that we are currently using. Conversion is dp = pt * (96 / 72).

This is not only a cosmetic change: text was unnecessarily small compared to GTK apps.

I went for a hardcoded value, rounding two decimals, but we could obvs do the calculation with a function:

```dart
double ptToPx(double points) {
  return points * 96 / 72;
}
```

## Unnecessary letter spacing

Current styles seem to be inheriting some letter spacing from Material/Flutter. This letter spacing is unnecessary and inconsistent with Yaru GTK.

I checked with @lyubomir-popov, visual designer at Canonical heavily involved in the development of Ubuntu Sans. Letter spacing is only acceptable for small caps or uppercase. My fix sets it to zero for all font styles. This results in a slightly higher density, good in some instances such as the sidebar (see https://github.com/ubuntu/yaru.dart/pull/1035)

| Before | After |
| --------|-------|
| Screenshot From 2025-09-30 13-20-52 |Screenshot From 2025-09-30 13-19-56 |

Related: #995

<!-- REMINDER:

1) For a bug fix, please target the `release` branch, else target `main`.

2) If this PR introduces any visual changes, please run:
```
flutter test --update-goldens
```
and commit the changes **or** if not covered by tests, attach screenshots below:

| Before | After |
|--------|-------|
|        |       |


Feichtmeier:
@juanruitina thanks for your work 🧑‍💻 👨🏻‍🎨 

You migh also want to check @Muqtxdir 'PR #996

Jupi007 (approved):
LGTM 👍 

The letter spacing is clearly visible between the "tofu" on the previous version of the golden images. I've always found that strange, I didn't think it was related to letter spacing 😅

PR #1039, opened 2025-10-16
fix: Use square corners for window controls when maximized

PsychotherapistSam:
This PR fixes a visual inconsistency where the close button on windows retains rounded corners even when the window is maximized.

The fix ensures that window controls use square corners in maximized state, matching the behavior of native platform conventions on windows.

_Visual Comparison_

Before (rounded corners when maximized):
Current Yaru close button with rounded corners

After (square corners when maximized):
Windows close button with square corners when maximized

I wasn't able to regenerate the golden test images due to font-related issues on my system but I hope this change is minimal enough (and not covered by the tests I think)


Feichtmeier (approved):
hi thanks! :) Glad that there are some windows users using this package!

PR #1042, opened 2025-11-04
feat(a11y): Add YaruFocusBorder widget

ashuntu:
Adds a `YaruFocusBorder` widget which can wrap most children and provide it a focus border without having to define custom borders for widgets individually. Obviously, this type of thing should be added by Flutter eventually, but this is a much easier (and more aligned with our style goals) holdover until then.

This should be able to be used to wrap virtually any focus-able widget, including regular buttons, tiles, etc.

<details>
<summary>Examples of wrapping some Yaru widgets with this</summary>

| Widget | Image |
|--------|-------|
| YaruSwitchButton (primary) | image |
| YaruSwitchButton (dropdown) | image |
| YaruCheckboxButton | image |
| YaruIconButton | image image |
| YaruSelectableContainer | image |

</details>

Related https://github.com/ubuntu/yaru.dart/issues/979

### DIFF
--- a/lib/constants.dart
+++ b/lib/constants.dart
@@ -32,3 +32,6 @@ const kYaruTitleBarItemHeight = 34.0;
 
 /// The default icon size
 const kYaruIconSize = 20.0;
+
+/// The default border width for various Yaru widgets.
+const kYaruBorderWidth = 2.0;
--- a/lib/src/widgets/yaru_focus_border.dart
+++ b/lib/src/widgets/yaru_focus_border.dart
@@ -1,0 +1,132 @@
+import 'package:flutter/material.dart';
+import 'package:yaru/constants.dart';
+
+enum _YaruFocusBorderVariant { primary, secondary, onSurface }
+
+/// Padding between the border and child widget.
+class YaruFocusBorderPadding {
+  static const zero = EdgeInsetsGeometry.zero;
+  static const small = EdgeInsets.all(2);
+  static const medium = EdgeInsets.all(4);
+  static const large = EdgeInsets.all(6);
+}
+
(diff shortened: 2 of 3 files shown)


ashuntu:
cc @juanruitina

Feichtmeier:
that's a nice solution!

but shouldnt it somehow ignore the lower focus by default? i mean if its a button and you focus tab cycle it will probably switch to the child focus then?

ashuntu:
> but shouldnt it somehow ignore the lower focus by default? i mean if its a button and you focus tab cycle it will probably switch to the child focus then?

I initially thought that as well, but it seems to not actually behave like that. With every widget I've tested, it behaves as you'd want it to and doesn't "double" the focus.

~~On a separate note, this does (by default) add a bit of padding around any widget within a `YaruFocusBorder`. I think I'll mark this as draft again to see if I can get something that doesn't impact widget size.~~ Fixed :)

Feichtmeier:
> > but shouldnt it somehow ignore the lower focus by default? i mean if its a button and you focus tab cycle it will probably switch to the child focus then?
> 
> I initially thought that as well, but it seems to not actually behave like that. With every widget I've tested, it behaves as you'd want it to and doesn't "double" the focus.
> 
> ~~On a separate note, this does (by default) add a bit of padding around any widget within a `YaruFocusBorder`. I think I'll mark this as draft again to see if I can get something that doesn't impact widget size.~~ Fixed :)

Okay 👍 
Let's get this in 

((( Btw YaruIconButton would be deprecated as we can do the same with the same to normal icon buttons )))

Feichtmeier (approved):

PR #1043, opened 2025-11-05
fix(a11y): Thicken borders around text inputs

ashuntu:
Another small adjustment: thicken the input border from 1 to 2.

| Existing | This PR |
|----------|---------|
| image image | image image |

Related https://github.com/ubuntu/yaru.dart/issues/979
> In form fields, outline should replace existing borders (2px, inside).

### DIFF
--- a/lib/src/themes/common_themes.dart
+++ b/lib/src/themes/common_themes.dart
@@ -39,7 +39,7 @@ AppBarTheme _createAppBarTheme(ColorScheme colorScheme) {
 
 InputDecorationTheme _createInputDecorationTheme(ColorScheme colorScheme) {
   final radius = BorderRadius.circular(kYaruButtonRadius);
-  const width = 1.0;
+  const width = kYaruBorderWidth;
   const strokeAlign = 0.0;
   final fill = colorScheme.surface.scale(
     lightness: colorScheme.isLight ? -0.05 : -0.1,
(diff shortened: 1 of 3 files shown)


lib/src/themes/common_themes.dart
```diff
@@ -39,7 +39,7 @@ AppBarTheme _createAppBarTheme(ColorScheme colorScheme) {
 
 InputDecorationTheme _createInputDecorationTheme(ColorScheme colorScheme) {
   final radius = BorderRadius.circular(kYaruButtonRadius);
-  const width = 1.0;
+  const width = kYaruBorderWidth;
```
Feichtmeier: The constant name here could be misleading. Where else do we use it? Maybe just create a new variable because borders are not that thick elsewhere

Feichtmeier (changes requested):

lib/src/themes/common_themes.dart
ashuntu: I introduced it in https://github.com/ubuntu/yaru.dart/pull/1042 since all focus-related borders should be the same width. Maybe we could do `kYaruFocusBorderWidth` or something.

Feichtmeier: Yes ☺️ that sounds better

Feichtmeier (approved):

PR #1044, opened 2025-11-10
fix(example): correct code snippet url in draggable page example

Timo-Schroeder:
In the example application, the code snippet for the draggable page did not load.
image

I fixed the typo, so the snippet shows up again.
image


Feichtmeier (approved):
Thank you!

PR #1045, opened 2025-11-14
feat(a11y): Add focus border to focusable Yaru widgets

### LINKED ISSUE #979: Add focus ring to all components for improved accessibility

ashuntu:
(some context in this comment: https://github.com/ubuntu/yaru.dart/issues/979#issuecomment-3486102612)

This adds a `YaruFocusBorder` wrapper around most Yaru widgets, which can be disabled by passing `hasFocusBorder: false` to the widget. Alternatively, we might be able to do some magic with a custom theme to modify the focus borders, but I'll leave that up for discussion on whether it's worth it or not since there are pros and cons to both approaches.

I believe this will resolve most focus-related issues while we wait for official Flutter focus borders (and these are likely more in-line with our design goals, anyway). Anything that isn't a Yaru widget in a downstream app can use the exact same `YaruFocusBorder` to wrap any custom widgets.

I split each commit based on the widgets they modify. I'd recommend pulling this down and running the example app yourself when reviewing to see how it looks.

See https://github.com/ubuntu/yaru.dart/pull/1042 for some example images. I understand that `YaruIconButton` may be deprecated in the future, but I included it in these changes for now.

If I missed any widgets or a different approach is desired, please let me know!

Fixes https://github.com/ubuntu/yaru.dart/issues/979


Feichtmeier:
damn if you would have created the branch in this repo I could have ran the publish action to see how it looks :P 
gonna take one or two evenings to find time to check
and thank you very much again for contributing!!! 

meanwhile maybe @Jupi007 could check

@ashuntu one optional thing that might be cool would be to set the defaultbehaviour by the YaruTheme widget so if one wants to disable this it could be done globally instead of in each widget

Jupi007 (approved):
I've just taken a look and it looks really good!

ashuntu:
> @ashuntu one optional thing that might be cool would be to set the defaultbehaviour by the YaruTheme widget so if one wants to disable this it could be done globally instead of in each widget

I added this in https://github.com/ubuntu/yaru.dart/pull/1045/commits/26c25ebd6bf3ab3d14adddf3b5a30687923960bd, let me know if that's what you were thinking.

Feichtmeier:
> > @ashuntu one optional thing that might be cool would be to set the defaultbehaviour by the YaruTheme widget so if one wants to disable this it could be done globally instead of in each widget
> 
> I added this in [26c25eb](https://github.com/ubuntu/yaru.dart/commit/26c25ebd6bf3ab3d14adddf3b5a30687923960bd), let me know if that's what you were thinking.

yes sorry for taking so long! amazing! tyvm!

PR #1048, opened 2025-12-05
feat: make theme and darkTheme non-nullable

adil192:
Makes `YaruThemeData.theme` and `YaruThemeData.darkTheme` return a `ThemeData` instead of `ThemeData?`.
No changes were made other than removing the two question marks.

This means users of the package don't have to deal with null values, e.g.:
```patch
 return YaruTheme(
   builder: (context, yaru, _) {
     return MaterialApp(
       key: _appKey,
-      theme: yaru.theme ?? yaruLight,
-      darkTheme: yaru.darkTheme ?? yaruDark,
+      theme: yaru.theme,
+      darkTheme: yaru.darkTheme,
       debugShowCheckedModeBanner: false,
       home: const ConnectPage(),
     );
   },
 );
```

### DIFF
--- a/lib/src/settings/inherited_theme.dart
+++ b/lib/src/settings/inherited_theme.dart
@@ -333,11 +333,11 @@ class YaruThemeData with Diagnosticable {
 
   /// The light theme of [variant] (or [yaruLight] if not available) merged with
   /// the `YaruThemeData` overrides.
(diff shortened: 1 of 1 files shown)


Feichtmeier (approved):
thanks, i really wonder why it was nullable ... : D

PR #1051, opened 2025-12-09
fix: Split button spacing

### LINKED ISSUE #1047: fix: yaru split button now has too much space
Image

Since the recent changes to focus and the focus rings there is now a rather huge gap in yaru split button
this should not look like this :) there should not be a gap at all

CC @ashuntu

ashuntu:
Reduces the space between filled split buttons to 1 and removes the gap between outlined split buttons.

See the first row of https://github.com/ubuntu/yaru.dart/issues/1047#issuecomment-3634719201 for what this looks like.

This also fixes an issue where the dropdown button was 1px smaller than the main button. At first this was a simple size change, but I think the solution now is more robust (https://github.com/ubuntu/yaru.dart/pull/1051/commits/80a0bb0a50b2eb54cc7d62016dff48dd2f7614a9). It should handle larger/smaller main button sizes gracefully, like so:

image

A button with unstyled text (or smaller) still has a square dropdown button as before, but if the size is increased, the dropdown button will just increase in height instead of staying the same. 

Fixes https://github.com/ubuntu/yaru.dart/issues/1047

---

UDENG-8681

### DIFF
--- a/lib/src/widgets/yaru_split_button.dart
+++ b/lib/src/widgets/yaru_split_button.dart
@@ -63,7 +63,7 @@ class YaruSplitButton extends StatelessWidget {
 
     final dropdownShape = switch (_variant) {
       _YaruSplitButtonVariant.outlined => NonUniformRoundedRectangleBorder(
-        hideLeftSide: false,
+        hideLeftSide: true,
         borderRadius: BorderRadius.horizontal(
           right: defaultRadius,
           left: Radius.zero,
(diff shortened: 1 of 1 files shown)


ashuntu:
~~*Draft for now because I'm not satisfied with just increasing the dropdown button size. We need a button solution in case the button content is a non-default text size.*~~ Done: https://github.com/ubuntu/yaru.dart/pull/1051/commits/80a0bb0a50b2eb54cc7d62016dff48dd2f7614a9

Feichtmeier (approved):
Awesome, thank you!

juanruitina:
Nice job 👏🏻

PR #1054, opened 2025-12-16
fix: Prevent YaruSplitButton from expanding infinitely

### LINKED ISSUE #1053: fix(Example): Web example shows a blank split button page
> oh no, now they are invisible in the xample https://ubuntu.github.io/yaru.dart/ :D 
> I guess it is an exception?
> 
> @ashuntu if you would clone this repo and work with this on your computer rather than the fork of you, we could just run this workflow here in pull requests to test it before merging:
> 
> Image 

 _Originally posted by @Feichtmeier in [#1047](https://github.com/ubuntu/yaru.dart/issues/1047#issuecomment-3636282598)_

ashuntu:
Also adds some very simple tests for YaruSplitButtons since there were none before.

This should be visually the same as https://github.com/ubuntu/yaru.dart/pull/1051, but it should fix issues where the buttons would infinitely expand and cause a render issue.

Fixes https://github.com/ubuntu/yaru.dart/issues/1053

---

UDENG-8681


Feichtmeier (approved):
THANKS! Works!

PR #1055, opened 2026-01-09
fix: Respect focus border settings for toggle buttons

ashuntu:
Previously toggle button focus borders were by default on and ignore the YaruTheme, now they should respect the overall YaruTheme (which, by default, is on).

---

UDENG-8728


Feichtmeier:
hi @ashuntu thanks! Looks good to me on a first glance
I would like a doubel check from @Jupi007 since he made those widgets, cheers and happy new year

Jupi007 (approved):
LGTM 👍 
Thanks @ashuntu and happy new year!

PR #1056, opened 2026-02-04
chore: create issue templates

deandreamatias:
Create GitHub issue templates inspired on [App Center](https://github.com/ubuntu/app-center) and [Ubuntu Desktop Provision](https://github.com/canonical/ubuntu-desktop-provision)


.github/ISSUE_TEMPLATE/config.yml
```diff
@@ -0,0 +1,6 @@
+blank_issues_enabled: true
+contact_links:
+  - name: Contact Canonical
+    url: https://ubuntu.com/contact-us/form
```
Feichtmeier, line 4: @ashuntu could you check if this is the correct contact url?

Feichtmeier (commented):
TYVM! Indeed good to have those

.github/ISSUE_TEMPLATE/config.yml
ashuntu: This is consistent with what we have in our repositories.

Although, it seems the file needs to be named with the `yml` extension (not `yaml`) for it to work, at least in my experience.

.github/ISSUE_TEMPLATE/config.yml
```diff
@@ -0,0 +1,6 @@
+blank_issues_enabled: true
```
ashuntu, line 1: I think this file needs to be named `config.yml` for it to be picked up by GitHub, not sure why.

ashuntu (changes requested):

.github/ISSUE_TEMPLATE/config.yml
deandreamatias: On official docs `yml` is use but by my experience `yaml` also works ([example](https://github.com/ubuntu/app-center/tree/main/.github/ISSUE_TEMPLATE)).
I will rename to follow official guidelines

.github/ISSUE_TEMPLATE/config.yml
deandreamatias: Updated

.github/ISSUE_TEMPLATE/config.yml
ashuntu: When I click the new issue button in the App Center I do not get the contact link as an option, but if I rename the config to `.yml` in a fork, it works.

Feichtmeier (approved):

Issue #1057, opened 2026-02-05
List Tile Overhaul Proposal


ashuntu:
In our apps, we often need to create list tiles with a similar style and we keep these custom widgets in each app. It would be useful to use Yaru instead to provide these tiles to us. Yaru already does this though the `YaruTile` widget and `Yaru[Switch/Radio/Checkbox]ListTile` widgets, but I would like to expand on these, pending approval:

## 1. Add a new `YaruListTile` widget

*or, replace the `YaruTile` widget with this one.*

This widget would be similar in look to a Material `ListTile` and `YaruTile`, but offer more flexibility and would not be an actual `ListTile` widget. At the moment, it is quite hard to achieve our design goals with Material `ListTiles` alone, and often difficult to add more "auxiliary" widgets like multiple leading/trailing widgets (see 2).

You can find an example of what the code would look like in the Security Center: https://github.com/canonical/desktop-security-center/blob/main/packages/security_center/lib/widgets/security_center_list_tile.dart. Though, I would plan on expanding on this to make it more versatile.

This would not be a breaking change.

## 2. Refactor `Yaru[Switch/Radio/Checkbox]ListTile`s with the new `YaruListTile` 

To meet our design goals, we'd like these tiles to be more flexible and support additional auxiliary widgets. For example, it would be useful to be able to add widgets on either side of the toggle buttons:

<img width="580" height="291" alt="Image" src="https://github.com/user-attachments/assets/1689dd24-e0b6-4f2e-827d-862ec29a6fcd" />

I believe this is impossible with the current setup without adding our own custom tiles.

This would also make these tiles meet design requirements for padding, font size, etc.

This would be a breaking change.

## 3. Add a `YaruTileList` widget

In addition to tiles, we often display them in bordered lists:

<img width="580" height="291" alt="Image" src="https://github.com/user-attachments/assets/60bb8cbc-71e3-40b1-8ab8-a583f4805a00" />

Or, alone but for navigation:

<img width="574" height="96" alt="Image" src="https://github.com/user-attachments/assets/f461c226-ad27-479e-9981-9feb6f09c058" />

At the moment, we do this manually in most places by creating a `YaruBorderContainer` (or similar) and adding `Divider` widgets.

I would propose we copy something similar to https://github.com/canonical/desktop-security-center/blob/main/packages/security_center/lib/widgets/tile_list.dart into Yaru for use in other Ubuntu Flutter apps.

This would not be a breaking change.

---

[...]

PR #1058, opened 2026-02-06
feat!: YaruInfoBox visual improvements

ashuntu:
Adjusts the look of YaruInfoTiles. Notably it makes the text sizes all `bodyMedium`, aligns the icon to the top left, and adjusts some of the padding.

It also removes the inner `ListTile` since it was causing more issues than its worth to achieve our style goals. This is a breaking change since it removes the `isThreeLine` field, but that field isn't really relevant without the underlying `ListTile`. This could potentially be included in https://github.com/ubuntu/yaru.dart/issues/1057 when a new tile is added, but for now I think this stands on its own.

Also adds some tests for this widget as there were none before.

|       | Before | After |
|-------|--------|-------|
| Light | image | image |
| Dark  | image | image |

cc @juanruitina 

---

UDENG-8485


Feichtmeier:
That looks indeed better! Thank you

If you changed the fields of the class I consider this a breaking change and you need to insert the exclamation ❗ after feat in the title/commit so the release bot bumps a major version

lib/src/widgets/yaru_info.dart
```diff
@@ -104,28 +103,21 @@ class YaruInfoBox extends StatelessWidget {
...
 
-  /// An optional trailing [Widget] inserted into the internal [ListTile].
+  /// An optional trailing [Widget] inserted to the right of the title.
   final Widget? trailing;
 
-  ///  Whether the internal [ListTile] is intended to display three lines of text.
-  /// If true, then [subtitle] must be non-null (since it is expected to give the second and third lines of text).
-  /// If false, the list tile is treated as having one line if the subtitle is null and treated as having two lines if the subtitle is non-null.
-  /// When using a [Text] widget for [title] and [subtitle], you can enforce line limits using [Text.maxLines].
-  final bool isThreeLine;
```
Feichtmeier, line 127: Small but Breaking change :)

PR #1060, opened 2026-02-09
fix: Reduce default ListTile text size

ashuntu:
This impacts all Material `ListTile` based widgets which includes `YaruMasterTile` and the toggle button list tiles (in the future, this will only impact `YaruMasterTile`).

There's other changes to make to the list tiles as described in https://github.com/ubuntu/yaru.dart/issues/1057, but this will align regular Material `ListTile`s with the rest of them.

Note the navigation tile size changes and the `YaruSwitchListTile` text sizes:

|       | Before | After |
|-------|--------|-------|
| Light | image | image |
| Dark  | image | image |

cc @juanruitina 

---

UDENG-8918


Feichtmeier (approved):

juanruitina:
Visual feedback for this:

- Font weight is now medium instead of regular. It should be regular. I think we need to change the style here, I don't think there's a good reason for labels to have a different font weight: https://github.com/ubuntu/yaru.dart/blob/f5929f4841536af3477d9144121164e49ac8b9af/lib/src/themes/text_theme.dart#L67
- There's some misalignment now for menu items, see below. Interesting to see the text shift down a couple pixels when making the text smaller, why is that?

image

- Title and subtitle look very, very close. I wonder if it has to do with line height, what is it for labelLarge and labelMedium, is it `auto`?

BTW, I see in here that we still refer to the Ubuntu font, should it be replaced with Ubuntu Sans? https://github.com/ubuntu/yaru.dart/blob/f5929f4841536af3477d9144121164e49ac8b9af/lib/src/themes/text_theme.dart#L89

Feichtmeier:
@ashuntu just add @juanruitina to the reviewer next time you need the ack by him. If you want to let him check the result you can create a new github pages deployment from actions, if this okay for you that is 😸 
image

Thanks everyone for your work on yaru.dart ! Highly appreciated/needed

ashuntu:
> Font weight is now medium instead of regular. It should be regular. I think we need to change the style here, I don't think there's a good reason for labels to have a different font weight

https://github.com/ubuntu/yaru.dart/pull/1061

> There's some misalignment now for menu items, see below. Interesting to see the text shift down a couple pixels when making the text smaller, why is that?

Sizing Material `ListTile`s has proved a little annoying because they size things in a very particular way. I believe we can adjust the density of the tile theme to account for some of this.

Keep in mind we can rework widgets to not depend on `ListTile`, but that's not part of this PR.

> Title and subtitle look very, very close. I wonder if it has to do with line height, what is it for labelLarge and labelMedium, is it auto?

It seems to be whatever the default is. `bodyMedium` has a height of 1.43, and `labelMedium` has a height of 1.33 (these are in terms of a multiple of the font size).

PR #1064, opened 2026-02-24
feat!: Adjust toggleables to new styles

ashuntu:
This updates toggleable widgets (checkboxes, switches, radios) to new designs.

***This is a breaking change for the toggle tiles as some of the fields were `ListTile`-specific.** If we'd rather go the deprecation route, we can do that but I'm not sure what to name the new tiles in that case.*

Summary of changes:
- Radio
  - Thicken unchecked border
  - Increase contrast on unchecked border in light mode
- Checkbox
  - Thicken unchecked border
  - Increase contrast on unchecked border in light mode
- Switch
  - Reduce switch size
  - Slightly reduce thumb size
- Toggleable buttons
  - Reduce text size
- Toggleable tiles
  - Base the list tiles on `YaruListTile` instead of `ListTile`
    - Removes `isThreeLine`, `dense`, `tileColor`, `selectTileColor`, `selected`, 
`visualDensity` fields
  - Tiles can now optionally accept a `control` to supply a custom control
    - Examples of this have been added to the relevant pages
  - Checkbox tiles now have the checkbox on the left instead of right

I believe this is the last remaining items for https://github.com/ubuntu/yaru.dart/issues/1057, but will confirm

| Before | After |
|--------|-------|
| image image image | image image image |

---

UDENG-9090
UDENG-8482
UDENG-8483
UDENG-8484


juanruitina:
Great visual updates. Some comments:

- [x] Same as radio buttons, checkboxes should also be on the left of the tile. Only switches make sense on the right.
- [x] The info tooltip, if any, should always be on the right.
- [x] To keep consistency with GTK Yaru, the chevron used for moving to a different page should be `YaruIcons.go_next`
- [ ] On focus, remove gray background, focus ring is sufficient.
- [ ] When focusing on an interactive element within a tile, only that element should have the focus ring, not the tile itself
- [x] And a very minimal visual tweak: the thumb in the switch seems to be 21px high. To match GNOME, it should be 20px high, with 3px of spacing all around.

Honestly, I don't think we should be using tiles for both checkboxes _and_ moving to a different page, switches are more common for that, so perhaps I'd drop that example.

About focus behaviour, I'll suggest that, by default, focus on a tile should trigger the toggleable, then focus can shift to interactive secondary elements like an information tooltip or the button to go to the next page. The current behaviour where focus moves first to the tile and then to the toggleable seems redundant. Screen readers should announce the tile is toggleable, and any secondary elements should have a good label and role.

ashuntu:
@juanruitina 

Let's reserve your 4th and 5th points for a different PR in the future. The rest of your points should be resolved, the screenshots are updated.

I'm not sure if there's any example of buttons around radio or checkboxes as there are for switches, if there are, I can add those to the examples. For now, I've removed those for radios/checkboxes examples since they aren't a flow we want to encourage.

juanruitina:
Looks good to me! Let's handle the focus changes in a separate PR then.

Jupi007 (approved):
LGTM 👍 
Very nice, thank you!

PR #1068, opened 2026-03-17
feat: Add option to hide the default popup icon

ashuntu:
This is useful to be able to display *just* an icon (or anything else) as the popup menu button.

| Before | After |
|--------|-------|
| image | image |

---

UDENG-9432


Feichtmeier:
That is a nice idea but since you want to hide the arrow maybe make the Boolean rather "include arrow = true"?

Feichtmeier (approved):

PR #1070, opened 2026-03-23
fix: Hide focus borders for nested list tile focuses

ashuntu:
Hides list tile focus borders when one if its children is focused, and not the tile. This is more consistent with GTK behavior.

| Before | After |
|--------|-------|
| image image | image image |

---

UDENG-9317

### DIFF
--- a/lib/src/widgets/yaru_radio_list_tile.dart
+++ b/lib/src/widgets/yaru_radio_list_tile.dart
@@ -74,6 +74,14 @@ class _YaruRadioListTileState<T> extends State<YaruRadioListTile<T>> {
   }
 
   @override
+  void dispose() {
+    super.dispose();
+    if (widget.focusNode == null) {
+      _focusNode.dispose();
+    }
+  }
+
+  @override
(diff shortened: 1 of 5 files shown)


lib/src/widgets/yaru_radio_list_tile.dart
```diff
@@ -59,52 +59,72 @@ class YaruRadioListTile<T> extends YaruToggleListTile {
     }
   }
 
+  @override
+  State<StatefulWidget> createState() => _YaruRadioListTileState<T>();
+}
+
+class _YaruRadioListTileState<T> extends State<YaruRadioListTile<T>> {
+  bool _tileHasFocus = false;
+  late final FocusNode _focusNode;
```
Feichtmeier, line 68: maybe we want to dispose the focus node here?

https://api.flutter.dev/flutter/widgets/FocusNode/dispose.html

Feichtmeier (changes requested):
minor thing to change, thanks for the help

Feichtmeier (approved):

PR #1073, opened 2026-04-02
fix: Minimize YaruSplitButton width

ashuntu:
Sets the main axis size for split buttons to `MainAxisSize.min` so that they don't take up unnecessary space.

---

UDENG-9615


Feichtmeier (approved):
ty, just tested on github pages!

Issue #905, opened 2024-07-08
How to change font family and font weight


kawsaramin101:
I have this code to change font family and font weight when using Yaru theme 

```
import 'package:flutter/material.dart';

import 'package:yaru/yaru.dart';

void main() async {
  await YaruWindowTitleBar.ensureInitialized();
  runApp(MainWidget());
}

class MainWidget extends StatelessWidget {
 
  const MainWidget({super.key});


  @override
  Widget build(BuildContext context) {
    return YaruTheme(
        data: const YaruThemeData(
            themeMode: ThemeMode.dark, useMaterial3: true),
        builder: (context, yaru, child) {
          final ThemeData lightTheme = yaru.theme ?? ThemeData.light();
          final ThemeData darkTheme = yaru.darkTheme ?? ThemeData.dark();
    
          return MaterialApp(
            home: const BaseLayout(),
            theme: _buildTheme(lightTheme, Brightness.light),
            darkTheme: _buildTheme(darkTheme, Brightness.light),
            themeMode: ThemeMode.dark,
            debugShowCheckedModeBanner: false,
            builder: (context, child) {
              return MediaQuery(
                data: MediaQuery.of(context).copyWith(
                  textScaler: const TextScaler.linear(1.0),
                ),
                child: child!,
              );
            },
          );
        });
  }
}

ThemeData _buildTheme(ThemeData base, Brightness brightness) {
  const FontWeight fontWeight = FontWeight.w300;
  const String fontFamily = 'RobotoMono';

  return base.copyWith(
      brightness: brightness,
      splashFactory: NoSplash.splashFactory,
      textTheme: const TextTheme(
        displayLarge: TextStyle(fontFamily: fontFamily, fontWeight: fontWeight),
        displayMedium:
            TextStyle(fontFamily: fontFamily, fontWeight: fontWeight),
        displaySmall: TextStyle(fontFamily: fontFamily, fontWeight: fontWeight),
        headlineLarge:
            TextStyle(fontFamily: fontFamily, fontWeight: fontWeight),
        headlineMedium:
            TextStyle(fontFamily: fontFamily, fontWeight: fontWeight),
        headlineSmall:
            TextStyle(fontFamily: fontFamily, fontWeight: fontWeight),
        titleLarge: TextStyle(fontFamily: fontFamily, fontWeight: fontWeight),
        titleMedium: TextStyle(fontFamily: fontFamily, fontWeight: fontWeight),
[...]

PR #1074, opened 2026-04-04
fix: remove `TextStyle.package` to allow `copyWith` to change the font

adil192:
This package's use of `TextStyle.package` causes problems when apps don't want to use the bundled Ubuntu font.

### Problem

The ideal approach to change the font would be something like
```dart
textStyle = textStyle.copyWith(fontFamily: 'MyNonYaruFont', package: null);
```
But the implementation of copyWith cannot remove textStyle.package, leading to the new fontFamily being `packages/yaru/MyNonYaruFont`.

Users of this package currently need to use workarounds like:
1. Make a function similar to `copyWith` that explicitly removes the `package` field:
   [ 🔗](https://github.com/adil192/no_more_background/blob/201e3ab3ebfda19953d5419bb6c0d80b2f24d9b4/lib/main.dart#L181-L225)
   But not only is this verbose, but it's likely to break in time as new fields are added to TextStyle.

2. Ditch Yaru's TextTheme completely, and use the default Flutter TextTheme:
   [ 🔗](https://github.com/saber-notes/saber/blob/4088ad373c4a1ac83a752aaabb00e0dc58e9a2a1/lib/components/theming/saber_theme.dart#L134-L154)
   But of course, we then lose Yaru's TextTheme :(.
   Also see https://github.com/ubuntu/yaru.dart/issues/905.

### Solution

This PR simply replaces `TextStyle(fontFamily: 'Ubuntu', package: 'yaru')` with `TextStyle(fontFamily: 'packages/yaru/Ubuntu')` which is functionally identical except resolving the issue stated above.
This lets apps choose their own fonts without messy boilerplate.

Unit tests have been added for this problem.

### DIFF
--- a/example/lib/code_snippet_button.dart
+++ b/example/lib/code_snippet_button.dart
@@ -110,9 +110,8 @@ class _CodeDialogState extends State<_CodeDialog> {
                         : vsTheme,
                     padding: const EdgeInsets.all(12),
                     textStyle: const TextStyle(
-                      fontFamily: 'UbuntuMono',
+                      fontFamily: 'packages/yaru/UbuntuMono',
                       fontSize: 16,
-                      package: 'yaru',
                     ),
                   ),
                 );
--- a/lib/src/themes/common_themes.dart
+++ b/lib/src/themes/common_themes.dart
@@ -7,7 +7,7 @@ import 'text_theme.dart';
 
 // AppBar
 
-AppBarTheme _createAppBarTheme(ColorScheme colorScheme) {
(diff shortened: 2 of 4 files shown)


adil192:
I need this PR to fix my app's i18n support so I'm going to switch my app to my own fork of yaru.dart for now.

To make maintenance easier, I'll combine this with #1075 so I only have one commit that I need to keep rebasing.

Feichtmeier:
@adil192 hey sorry I missed this, we can get this in, I suppose you only need to merge in master? sorry!

Edit: alright, waiting for your other PR then!

PR #1075, opened 2026-04-04
feat: accept fontFamily parameter in createYaruTheme

adil192:
This PR adds font parameters to `createYaruTheme`, `createYaruLightTheme`, and `createYaruDarkTheme`.

When the theme is created, its text theme is passed around to several `...ThemeData` classes, making it difficult to completely override the font later.
Adding the parameters right at the start fixes this issue.
This has no impact on YaruVariant.

I had to workaround this recently in https://github.com/saber-notes/saber/commit/a4531475c4b2cd0c28ed2ad7ccca2478b98248ef. Note that changing the font isn't purely aesthetic, but is sometimes needed for language support.

### DIFF
--- a/lib/src/themes/common_themes.dart
+++ b/lib/src/themes/common_themes.dart
@@ -7,7 +7,7 @@ import 'text_theme.dart';
 
 // AppBar
 
-AppBarTheme _createAppBarTheme(ColorScheme colorScheme) {
+AppBarTheme _createAppBarTheme(ColorScheme colorScheme, TextTheme textTheme) {
   return AppBarTheme(
     shape: Border(
       bottom: BorderSide(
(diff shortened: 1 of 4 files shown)


Feichtmeier (approved):

lib/src/themes/common_themes.dart
```diff
@@ -7,7 +7,7 @@ import 'text_theme.dart';
 
 // AppBar
 
-AppBarTheme _createAppBarTheme(ColorScheme colorScheme) {
+AppBarTheme _createAppBarTheme(ColorScheme colorScheme, TextTheme textTheme) {
```
Feichtmeier: I nowadays (maybe not everywhere here in yaru) have the golgen rule that I make parameters named after more than one :) would be cool if you could do this here, too

Feichtmeier (changes requested):

lib/src/themes/common_themes.dart
adil192: No problem :)

Feichtmeier (approved):

PR #1077, opened 2026-04-23
fix(a11y): improve YaruAutocomplete accessibility and focus traversal

### LINKED ISSUE #1076: A11Y: improving autocomplete
Currently the autocomplete is pretty much unuseable for screen reader users. You can't tell what option you're selecting and you also can't tab past it as it's trapping you in the input.

paul-geoghegan:
fixes #1076 
Resolves focus traps, adds screen reader announcements via SemanticsService, and safely handles async search race conditions without breaking parent widget focus nodes.

There is still the outstanding issue of i18n as this pr only uses English strings currently so we would need to figure out how best to move forward before this fix can be fully implemented.

### DIFF
--- a/lib/src/widgets/yaru_autocomplete.dart
+++ b/lib/src/widgets/yaru_autocomplete.dart
@@ -1,5 +1,8 @@
+import 'dart:async';
 import 'package:flutter/material.dart';
 import 'package:flutter/scheduler.dart';
+import 'package:flutter/semantics.dart';
+import 'package:flutter/services.dart';
 
 import 'yaru_border_container.dart';
 
@@ -67,11 +70,25 @@ class YaruAutocomplete<T extends Object> extends StatefulWidget {
 class _YaruAutocompleteState<T extends Object>
     extends State<YaruAutocomplete<T>> {
   double? _optionsWidth;
+  bool _hasNavigated = false;
+  Iterable<T> _latestOptions = const [];
+  FocusNode? _internalFocusNode;
+  FocusOnKeyEventCallback? _originalOnKeyEvent;
+  Object? _activeSearchToken;
 
   @override
   void initState() {
     super.initState();
     _updateOptionsWidth();
+  }
+
+  @override
+  void dispose() {
+    _internalFocusNode?.removeListener(_onFocusChanged);
+    if (_internalFocusNode != null) {
+      _internalFocusNode!.onKeyEvent = _originalOnKeyEvent;
+    }
+    super.dispose();
   }
 
   @override
@@ -86,6 +103,7 @@ class _YaruAutocompleteState<T extends Object>
     _optionsWidth = widget.optionsWidth;
     if (_optionsWidth == null) {
       WidgetsBinding.instance.addPostFrameCallback((_) {
+        if (!mounted) return;
         final width = (context.findRenderObject() as RenderBox?)?.size.width;
         if (_optionsWidth != width) {
           setState(() => _optionsWidth = width);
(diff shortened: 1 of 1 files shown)


lib/src/widgets/yaru_autocomplete.dart
```diff
@@ -94,14 +112,112 @@ class _YaruAutocompleteState<T extends Object>
...
+    if (_internalFocusNode?.hasFocus == true) {
+      _hasNavigated = false;
+    }
+  }
+
+  void _announceResultsCount(int count, String text) {
+    if (text.isEmpty) {
+      SemanticsService.announce('Input cleared', TextDirection.ltr);
+    } else if (count == 0) {
+      SemanticsService.announce('No options found', TextDirection.ltr);
```
ashuntu, line 125: These strings should be able to be passed in to the `YaruAutocomplete` as optional fields. Perhaps also the announce text direction.

lib/src/widgets/yaru_autocomplete.dart
```diff
@@ -94,14 +112,112 @@ class _YaruAutocompleteState<T extends Object>
...
+    } else {
+      if (count == 1 && _latestOptions.isNotEmpty) {
+        if (widget.displayStringForOption(_latestOptions.first) == text) {
+          return;
+        }
+      }
+      final message = count == 1
+          ? '1 option available'
+          : '$count options available';
```
ashuntu, line 134: This text should also be able to be passed in as an optional field for `YaruAutocomplete`, maybe as a callback since the string needs to know about the result count. The translations we use in our apps have support for plurals, so in the ideal case we wouldn't need to do this special handling when using translated strings. Example: https://github.com/canonical/desktop-security-center/blob/main/packages/security_center/lib/l10n/app_en.arb#L35

lib/src/widgets/yaru_autocomplete.dart
```diff
@@ -94,14 +112,112 @@ class _YaruAutocompleteState<T extends Object>
...
+          _internalFocusNode?.onKeyEvent = (node, event) {
+            if (event is KeyDownEvent) {
+              if (event.logicalKey == LogicalKeyboardKey.arrowDown) {
+                if (!_hasNavigated) {
+                  _hasNavigated = true;
+                  if (_latestOptions.isNotEmpty) {
+                    final option = _latestOptions.first;
+                    SemanticsService.announce(
+                      widget.displayStringForOption(option),
+                      TextDirection.ltr,
```
ashuntu, line 189: We should use a `YaruAutocomplete` field for these TextDirections so they can be overridden.

ashuntu (changes requested):
Tested locally and it does seem to resolve a lot of the issues with the screen reader.

I know there's some ongoing discussions about translations. At a minimum I think the strings added should be overridable by the caller.

Also, make sure to bump `.fvmrc` and `pubspec.yaml` Flutter versions to at least `3.41.x`, since announcements don't work on Linux before then.

There's also some analyzer errors to resolve https://github.com/ubuntu/yaru.dart/actions/runs/24838240076/job/72717452809?pr=1077

paul-geoghegan:
@ashuntu thanks for testing it. I've rebased onto main.

I'll fix the analyzer failure. The logs had expired, but running it locally shows two lints (`omit_local_variable_types` and `avoid_types_on_closure_parameters`), so it's a quick fix.

On the strings, I'll stop waiting on the i18n discussion and do what you suggested:
- optional fields for "Input cleared" and "No options found", defaulting to the current English
- a `String Function(int count)` callback for the result count, so apps can use their own plural strings
- an optional `TextDirection`. If it isn't set, I'd use `Directionality.of(context)` instead of hardcoding ltr.

Let me know if you want different names or would rather have it all in one config object.

I can't do the Flutter bump without changing dev dependencies. `animated_vector_gen` (0.2.3 is the latest) pins `analyzer: ^6.2.0`, and from 3.41.0 `flutter_test` needs a `test` version that requires analyzer 8+. 3.38.0 still resolves, but 3.41.0, 3.41.9, 3.44.6 and 3.47.5 all fail `pub get`.

`dependency_overrides` doesn't get around it. Overriding analyzer lets `pub get` pass, but then `build_runner` fails to compile because `build_resolvers` 2.4.2 and `dart_style` 2.3.6 use APIs that were removed in analyzer 8. Bumping `build_resolvers` to 3.x needs `build` 4.x, and `animated_vector_gen` pins `build: ^2.4.1`.

As far as I can see, `animated_vector_gen` and `build_runner` are only used to regenerate `yaru_animated_icons.g.dart`. That file is already committed, and CI never runs `build_runner`. So the options I see are:
1. drop them from dev_dependencies in this PR
2. do the Flutter bump in its own PR and keep this one to the a11y changes
3. bump to 3.38 for now, though announcements still won't work on Linux with that

I'd go with 2, but I'm fine with whichever you prefer.

ashuntu (approved):
LGTM

@Jupi007 may have input, too. I think this is the first time we're introducing strings in this project in this way

ashuntu:
Like you say, we can do a Flutter upgrade in a separate PR

Jupi007 (approved):
As for the strings, I agree with @ashuntu's solution.

Regarding `animated_vector_gen`, I am currently updating its dependencies. But ofc if upgrading Flutter is urgent, you can remove `animated_vector_gen` from the dependencies.

paul-geoghegan:
Thanks both. Without the bump the announcements don't work on Linux, so I'd rather not wait. I'll open a PR that bumps Flutter to 3.41 and removes `animated_vector_gen` for now. It can go back in once your update is released.

dvd-bnc:
Hi! I updated the various libraries, they should work. Do note that v0.3.0 of annotations/gen is "breaking" because animated_vector_annotations doesn't re-export the whole library to avoid breaking build_runner, you just need to import the missing libraries which in most cases should be some dart:ui stuff and animated_vector. Sorry for the trouble again!

PR #1079, opened 2026-05-23
fix: wrap ListTile in Material to fix invisible ink splashes in YaruMasterTile

dongfengweixiao:
While syncing with [dongfengweixiao/starter](https://github.com/dongfengweixiao/starter), the following exception was thrown when tapping the hamburger button to open the sidebar:

```
══╡ EXCEPTION CAUGHT BY FLUTTER FRAMEWORK ╞═════════════════════════════════════════════════════════
The following assertion was thrown:
ListTile background color or ink splashes may be invisible.
The ListTile is wrapped in a DecoratedBox that has a background color. Because ListTile paints its
background and ink splashes on the nearest Material ancestor, this DecoratedBox will hide those
effects.
To fix this, wrap the ListTile in its own Material widget, or remove the background color from the
intermediate DecoratedBox.

ListTile: ListTile(onTap: Closure: () => void, selected: true)
DecoratedBox:
  DecoratedBox(bg: BoxDecoration(color: Color(alpha: 0.0400, red: 0.1098, green: 0.1098, blue: 0.1098, colorSpace: ColorSpace.sRGB), borderRadius:
  BorderRadius.circular(8.0)))
════════════════════════════════════════════════════════════════════════════════════════════════════
```


Feichtmeier:
Hi thanks! 

Only thing that comes to my mind that could have a problem with this are the recent focus rings of @ashuntu 
Let him double check

dongfengweixiao:
some test case failed, need i create new image?

ashuntu:
Focus rings still seem to work properly. I'm not sure if there's an example of this happening the Yaru example app; it would be helpful if there was to confirm this fix.

> some test case failed, need i create new image?

You'll need to regenerate golden images with `fvm flutter test --update-goldens` (and confirm the resulting images look correct).

Feichtmeier:
The goldens are Auto Regened at night

ashuntu (approved):

PR #1081, opened 2026-06-03
feat: add expanded option to YaruSplitButton

matthew-hagemann:
Add's an `expanded` flag (defaults false for backwards compatability) that allows for the main action button to fill the width given by its parent. This will help solve a an issue in the prompting client where it is difficult to get to YaruSplitButtons side by side with equal width if the text is different in length.

For example with the prompting client, it allows for:
### Before
Screenshot From 2026-06-03 16-45-35
Screenshot From 2026-06-03 16-55-36

### After
Screenshot From 2026-06-03 16-46-32
Screenshot From 2026-06-03 16-53-31

### DIFF
--- a/lib/src/widgets/yaru_split_button.dart
+++ b/lib/src/widgets/yaru_split_button.dart
@@ -16,6 +16,7 @@ class YaruSplitButton extends StatelessWidget {
     this.radius,
     this.menuWidth,
     this.hasFocusBorder,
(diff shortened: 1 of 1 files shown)


juanruitina:
This looks great, thank you for the work @matthew-hagemann. This should also work good for any modal we may need elsewhere.

Feichtmeier (approved):

PR #1086, opened 2026-07-22
fix: RTL search field

adil192:
Fixes the search field in RTL (e.g. Arabic) layouts. The search button now comes before the text, and no longer overlaps with the clear button.

Before and after with inactive search:

 

Before and after with active search:

 

Note that the cutoff "clear" button above is just an artifact of my goldens. It doesn't overflow in real life:
image


Feichtmeier:
Than you @adil192 !

adil192:
@Feichtmeier There are a few more places where RTL can be improved by replacing `EdgeInsets` with `EdgeInsetsDirectional` and `Alignment` with `AlignmentDirectional`. I didn't include them in this PR since I'm not familiar enough with Yaru to properly test them.

